Issue #4, opened 2018-06-27
Enhance the output formatting


xstefank:
Currently the list of issues is aggregated in the check description. Maybe add a separate line for each failure, or display only the last check for the ease of processing / repairing in steps.

PR #9, opened 2018-07-23
#4 Output formatting

khermano:
issue #4

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,45 @@
 # tyr
-
+  
 ![Powered by OpenShift Online](https://www.openshift.com/images/logos/powered_by_openshift.png)
 
 Pull request template status check
+
+## Running
+
+1. Clone Tyr - https://github.com/xstefank/tyr
+1. Build with Maven - mvn clean install
+1. Create a testing repository - for start you can fork my https://github.com/xstefank/test-repo where the PR template is set up or you can create your own :)
+    - For creating new repository check the github manuals/help
+    - PRs needs to be created to YOUR repository, you need to specify it under the base fork field in the Open a Pull Request page
+1. The project requires two properties to be set up to run correctly
+    - GitHub OAuth token - the setup will be described later
+    - Configuration file specifying the format of the PR - the example can be found in src/main/resources/format.yaml, there are two ways how to specify to the server where is this file located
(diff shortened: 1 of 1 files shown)


src/main/java/org/xstefank/check/TemplateChecker.java
```diff
@@ -31,22 +31,23 @@ public TemplateChecker() {
     }
 
     public void checkPR(JsonNode payload) {
-        StringJoiner joiner = new StringJoiner(", ");
+        //StringJoiner joiner = new StringJoiner(", ");
```
MMarus: Please do not commit commented code. In version control we can see what code was there before :).

src/main/java/org/xstefank/check/TemplateChecker.java
```diff
@@ -31,22 +31,23 @@ public TemplateChecker() {
...
             String message = check.check(payload);
             if (message != null) {
-                joiner.add(message);
+                description = message;
             }
         }
 
         log.info("updating status");
-        String description = joiner.toString();
+        //String description = joiner.toString();
```
MMarus: Please do not commit commented code. In version control we can see what code was there before :).

src/main/java/org/xstefank/check/TemplateChecker.java
```diff
@@ -31,22 +31,23 @@ public TemplateChecker() {
...
 
         GitHubAPI.updateCommitStatus(config.getRepository(),
                 payload.get(Utils.PULL_REQUEST).get(Utils.HEAD).get(Utils.SHA).asText(),
-                description.isEmpty() ? CommitStatus.SUCCESS : CommitStatus.ERROR,
+                description == "" ? CommitStatus.SUCCESS : CommitStatus.ERROR,
```
MMarus: This is unnecessary. It is better to use .isEmpty() as it was before.

src/main/java/org/xstefank/check/TemplateChecker.java
```diff
@@ -31,22 +31,23 @@ public TemplateChecker() {
...
                 config.getStatusUrl(),
-                description.isEmpty() ? "valid" : description, "PR format check");
+                description == "" ? "valid" : description, "PR format check");
```
MMarus: This is unnecessary. It is better to use .isEmpty() as it was before.

Issue #2, opened 2018-06-27
Add check for the commit title message


xstefank:
Including adding a format regex to the configuration file. Possibly allowing user to specify it must be same as PR title.

PR #12, opened 2018-07-24
#2 Added check for the commit title

lukasburda:
I just solved issue "Add check for the commit title message #2".

### DIFF
--- a/src/main/java/org/xstefank/check/TemplateChecker.java
+++ b/src/main/java/org/xstefank/check/TemplateChecker.java
@@ -31,16 +31,16 @@ public class TemplateChecker {
     }
 
     public void checkPR(JsonNode payload) {
-        StringJoiner joiner = new StringJoiner(", ");
+        String description = "";
         for (Check check : checks) {
             String message = check.check(payload);
             if (message != null) {
-                joiner.add(message);
+                description = message;
+                break;
             }
         }
 
         log.info("updating status");
-        String description = joiner.toString();
 
         GitHubAPI.updateCommitStatus(config.getRepository(),
                 payload.get(Utils.PULL_REQUEST).get(Utils.HEAD).get(Utils.SHA).asText(),


src/main/java/org/xstefank/model/Utils.java
```diff
@@ -8,47 +8,52 @@
...
-                    is.close();
-                }
-            } catch (IOException e) {
-                // intentionally ignored
-            }
-        }
-
-        return null;
-    }
+	public static final String GITHUB_BASE = "https://api.github.com";
```
xstefank: I am sorry, but please fix your IDE settings to not change the TAB size, this is really unnecessary. It is then really hard to catch what actually are your changes.

src/main/java/org/xstefank/model/yaml/Format.java
```diff
@@ -16,19 +17,15 @@ public void setTitle(String title) {
         this.title = title;
     }
 
-    public Description getDescription() {
-        return description;
-    }
+    public String getCommit() { return commit; }
```
xstefank: I don't have checkstyle set up (that's on me), but please do not use one line methods

src/main/java/org/xstefank/model/yaml/Format.java
```diff
@@ -16,19 +17,15 @@ public void setTitle(String title) {
...
-        this.description = description;
-    }
+    public void setCommit(String commit) { this.commit = commit; }
 
-    public List<String> getAdditional() {
-        return additional;
-    }
+    public Description getDescription() { return description; }
 
-    public void setAdditional(List<String> additional) {
```
xstefank: In general, try to avoid changing code that is not a part of the fix for a given issue. I understand that this code is not probably being used anywhere so IDE is doing its magic, but try to avoid it. This should be a separate fix for cleaning dead code.

src/main/java/org/xstefank/webhook/WebHookEndpoint.java
```diff
@@ -1,14 +1,15 @@
 package org.xstefank.webhook;
 
-import com.fasterxml.jackson.databind.JsonNode;
```
xstefank: Same, nothing bad, but please understand that it is not a part of the fix for this issue :) and it's only polluting the PR with non-relevant information (this is again IDE setting, you should be aware of what code are you changing)

src/main/resources/format.yaml
```diff
@@ -4,6 +4,7 @@ statusUrl: "https://github.com/xstefank/test-repo/blob/master/.github/pull_reque
 
 format:
   title: "\\[WFLY-\\d+\\]\\s+.*|WFLY-\\d+\\s+.*"
+  commit: "\\[WFLY-\\d+\\]\\s+.*|WFLY-\\d+\\s+.*"
```
xstefank: One additional question - do you think it would be possible to avoid copy-paste of the regex if title and commit messages are expected to be same? Can you come up with some solution to this problem?

lukasburda:
I fixed up your comments.

src/main/java/org/xstefank/check/TemplateChecker.java
```diff
@@ -3,13 +3,14 @@
...
+
 import org.jboss.logging.Logger;
 import org.xstefank.api.GitHubAPI;
 import org.xstefank.check.additional.AdditionalChecks;
 import org.xstefank.model.CommitStatus;
-import org.xstefank.model.yaml.FormatConfig;
-import org.xstefank.model.yaml.Format;
 import org.xstefank.model.Utils;
+import org.xstefank.model.yaml.Format;
+import org.xstefank.model.yaml.FormatConfig;
```
xstefank: The reorder of imports is not necessary

src/main/java/org/xstefank/check/TemplateChecker.java
```diff
@@ -61,6 +63,12 @@ public void checkPR(JsonNode payload) {
             checks.add(new RequiredRowsCheck(format.getDescription().getRequiredRows()));
         }
 
+        if (format.getCommit() != null) {
+            checks.add(new LatestCommitCheck(format.getCommit()));
+        } else if (format.getTitle() != null) {
+            checks.add(new LatestCommitCheck(format.getTitle()));
```
xstefank: the ```format.getTitle()``` is called too many times in this method, maybe you can refactor a little?

src/main/java/org/xstefank/model/yaml/Format.java
```diff
@@ -4,7 +4,7 @@
 
 public class Format {
 
-    private String title;
+    private String title, commit;
```
xstefank: Again, I need to setup checkstyle, but please use one line for definition of attributes

src/main/resources/format.yaml
```diff
@@ -4,6 +4,7 @@ statusUrl: "https://github.com/xstefank/test-repo/blob/master/.github/pull_reque
 
 format:
   title: "\\[WFLY-\\d+\\]\\s+.*|WFLY-\\d+\\s+.*"
+# commit: "regex"
```
xstefank: why this needs to be commented?

lukasburda: I think this format is used like a template or not? I want to show that you can set for commit different regex.

lukasburda: because by default it uses regex from title.

xstefank: so if you want to showcase this option, this line shouldn't be present, this comment doesn't add any value :)

xstefank:
@lukasburda it looks good now, rebase to the latest master to resolve the conflict please

xstefank:
@lukasburda please rebase to only one commit

PR #13, opened 2018-07-24
Fixing Output formatting

khermano:
Issue #4


src/main/java/org/xstefank/check/TemplateChecker.java
```diff
@@ -31,16 +31,15 @@ public TemplateChecker() {
     }
 
     public void checkPR(JsonNode payload) {
-        StringJoiner joiner = new StringJoiner(", ");
+        String description = "";
         for (Check check : checks) {
             String message = check.check(payload);
             if (message != null) {
-                joiner.add(message);
+                description = message;
```
xstefank: Thank you for the PR! This will be nice way of how to present information to the user, however, in this way the description will get rewritten with each failed check and only the last value will get propagated to the user. Is it intended? In other words, does the order of failed checks in which they are presented to the end user matter?

Issue #7, opened 2018-06-27
Add possibility to bypass PR template check


xstefank:
For instance by special string contained in the title that clearly states that the verification is not required

Additionally, it can be beneficial to allow users to state what message should be contained in title (or somewhere else) to indicate that the Tyr should not run

xstefank:
Adding sample format configuration:

```
skipCondition:
  title
  pattern: ".*"
```

PR #22, opened 2018-08-06
Add possibility to bypass PR template check

khermano:
Issue #7


src/main/java/org/xstefank/check/CheckSkip.java
```diff
@@ -0,0 +1,65 @@
...
+    private SkippingPatterns skippingPatterns;
+    private JsonNode payload;
+
+    public CheckSkip(SkippingPatterns skippingPatterns, JsonNode payload) {
+        this.payload = payload;
+        this.skippingPatterns = skippingPatterns;
+    }
+
+    public boolean shouldSkip() {
+        if (skippingPatterns != null && payload != null) {
```
xstefank: this should be checked in the constructor

src/main/java/org/xstefank/check/CheckSkip.java
```diff
@@ -0,0 +1,65 @@
...
+        }
+        return false;
+    }
+
+    private boolean skipByDescriptionFirstRow() {
+        if (skippingPatterns.getDescriptionPattern() != null) {
+            log.info("checking descriptionPattern in the first row of the description");
+            String description = payload.get(Utils.PULL_REQUEST).get(Utils.BODY).asText();
+            String firstRow = description;
+            if (description.contains(System.getProperty("line.separator"))) {
```
xstefank: there is definitely a nicer way of how you can implement this

src/main/java/org/xstefank/check/TemplateChecker.java
```diff
@@ -91,4 +93,12 @@ private static FormatConfig readConfig() {
         }
     }
 
+    private boolean shouldSkip(JsonNode payload) {
+        if (config.getFormat().getSkippingPatterns() != null) {
```
xstefank: you are checking both payload and skippingPatterns != null also in the call on line 99

src/main/java/org/xstefank/check/SkipChecker.java
```diff
@@ -0,0 +1,61 @@
...
+    private SkippingPatterns skippingPatterns;
+    private JsonNode payload;
+
+    public SkipChecker(SkippingPatterns skippingPatterns, JsonNode payload) {
+        this.payload = payload;
+        this.skippingPatterns = skippingPatterns;
+    }
+
+    public boolean shouldSkip() {
+        if (skippingPatterns != null && payload != null) {
```
xstefank: this should be checked in the constructor

src/main/java/org/xstefank/check/TemplateChecker.java
```diff
@@ -31,29 +31,31 @@ public TemplateChecker() {
     }
 
     public void checkPR(JsonNode payload) {
-        String description = "";
-        for (Check check : checks) {
-            String message = check.check(payload);
-            if (message != null) {
-                description = message;
-                break;
+        if (payload != null && !this.shouldSkip(payload)) {
```
xstefank: please do not indent whole content of the method if you can easily refactor it

xstefank: also `this` is not necessary here

src/main/resources/format.yaml
```diff
@@ -3,6 +3,10 @@ repository: "xstefank/test-repo"
 statusUrl: "https://github.com/xstefank/test-repo/blob/master/.github/pull_request_template.md"
 
 format:
+  skippingPatterns:
+    titlePattern: "((.*)?NO JIRA REQUIRED(.*)?)|([^do not|don't]\\sskip.*template.*check(.*)?)|([^do not|don't]\\sbypass.*template.*check(.*)?)"
```
xstefank: generally a really good PR :) one last minor comment, just a matter of taste, can you rename skippingPatterns to skipPatterns, and titlePattern -> title (and other two the same) please?

src/main/java/org/xstefank/check/SkipChecker.java
```diff
@@ -0,0 +1,61 @@
...
+        }
+        return false;
+    }
+
+    private boolean skipByDescriptionFirstRow() {
+        if (skippingPatterns.getDescriptionPattern() != null) {
+            log.info("checking descriptionPattern in the first row of the description");
+            String description = payload.get(Utils.PULL_REQUEST).get(Utils.BODY).asText();
+            String firstRow = description;
+            if (description.contains(System.getProperty("line.separator"))) {
```
xstefank: the must be a nicer way how you can implement this

src/main/java/org/xstefank/check/SkipChecker.java
```diff
@@ -0,0 +1,67 @@
...
+        if (payload != null ) {
+            this.payload = payload;
+        } else {
+            throw new IllegalArgumentException("Payload is null!");
+        }
+        if (skipPatterns != null) {
+            this.skipPatterns = skipPatterns;
+        } else {
+            throw new IllegalArgumentException("SkipPatterns is null!");
+        }
```
xstefank: Nothing wrong here now, but you should rather check if the params **are** null and in that case throw an exception. Probably in this case two separate if statements are redundant because you need both arguments to be present

src/main/java/org/xstefank/check/SkipChecker.java
```diff
@@ -0,0 +1,67 @@
...
+
+    }
+
+    public boolean shouldSkip() {
+        log.info("checking skipPatterns");
+        return (skipByTitle() || skipByCommit() || skipByDescriptionFirstRow());
```
xstefank: outer braces are redundant

src/main/java/org/xstefank/check/SkipChecker.java
```diff
@@ -0,0 +1,67 @@
...
+            return (latestCommitCheck.check(payload) == null);
+        }
+        return false;
+    }
+
+    private boolean skipByDescriptionFirstRow() {
+        if (skipPatterns.getDescription() != null) {
+            log.info("checking first row of the description");
+            String description = payload.get(Utils.PULL_REQUEST).get(Utils.BODY).asText();
+            String firstRow = description;
```
xstefank: still this line is hard to read for me, can you try to find another way please?

src/main/java/org/xstefank/check/TemplateChecker.java
```diff
@@ -31,29 +31,31 @@ public TemplateChecker() {
     }
 
     public void checkPR(JsonNode payload) {
-        String description = "";
-        for (Check check : checks) {
-            String message = check.check(payload);
-            if (message != null) {
-                description = message;
-                break;
+        if (!shouldSkip(payload)) {
```
xstefank: still indenting whole content of the method

src/main/java/org/xstefank/model/yaml/SkipPatterns.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.regex.Pattern;
+
+public class SkipPatterns {
+
+    private Pattern title;
+    private Pattern commit;
+    private Pattern description;
+
+    public Pattern getTitle() {
+
```
xstefank: unnecessary empty lines

src/main/resources/format.yaml
```diff
@@ -3,6 +3,10 @@ repository: "xstefank/test-repo"
 statusUrl: "https://github.com/xstefank/test-repo/blob/master/.github/pull_request_template.md"
 
 format:
+  skipPatterns:
+    title: "((.*)?NO JIRA REQUIRED(.*)\\s?)|([^do not|don't]\\sskip.*template.*check(.*)\\s?)|([^do not|don't]\\sbypass.*template.*check(.*)\\s?)"
+    commit: "((.*)?NO JIRA REQUIRED(.*)\\s?)|([^do not|don't]\\sskip.*template.*check(.*)\\s?)|([^do not|don't]\\sbypass.*template.*check(.*)\\s?)"
+    description: "((.*)?NO JIRA REQUIRED(.*)\\s?)|([^do not|don't]\\sskip.*template.*check(.*)\\s?)|([^do not|don't]\\sbypass.*template.*check(.*)\\s?)"
```
xstefank: `([^do not|don't]\\sskip.*template.*check(.*)\\s?)` -- if this regex will be matched the Tyr won't be included right? So `"do not skip template check"` included in title/commit/description will actually skip the template check??

src/main/java/org/xstefank/check/TemplateChecker.java
```diff
@@ -91,4 +93,9 @@ private static FormatConfig readConfig() {
         }
     }
 
+    private boolean shouldSkip(JsonNode payload) {
+        SkipChecker skipChecker = new SkipChecker(config.getFormat().getSkipPatterns(),payload);
+        return skipChecker.shouldSkip();
```
xstefank: I'll leave this to your judgement, do you think that the instance creation of the `SkipChecker` is necessary here? It will be created only for this one call and then eligible for garbage collection. Can you find some other way?

src/main/java/org/xstefank/check/TemplateChecker.java
```diff
@@ -91,4 +92,38 @@ private static FormatConfig readConfig() {
         }
     }
 
+    public boolean shouldSkip(JsonNode payload) {
```
xstefank: better, but still I think that this code is not particularly tied with TemplateChecker and should be moved to separate class

src/main/java/org/xstefank/webhook/WebHookEndpoint.java
```diff
@@ -20,7 +20,9 @@
     @Path("/pull-request")
     @Consumes(MediaType.APPLICATION_JSON)
     public void processPullRequest(JsonNode pullRequestPayload) {
-        templateChecker.checkPR(pullRequestPayload);
+        if (templateChecker.shouldSkip(pullRequestPayload)) {
```
xstefank: definitely supports my previous comment, it should not be required to call these methods (shouldSkip and checkPR) separately on the same object

Issue #17, opened 2018-07-26
Allow to skip commit check with title defined


xstefank:
The current inclusion of the LatestCommitCheck depends on the included title message in the format configuration. It should be possible to include title check without the need to also check for some general regex for commit.

PR #24, opened 2018-08-20
[Issue #17] Allow to skip commit check with title defined

lukasburda:
Issue: https://github.com/xstefank/tyr/issues/17

### DIFF
--- a/src/main/java/org/xstefank/check/TemplateChecker.java
+++ b/src/main/java/org/xstefank/check/TemplateChecker.java
@@ -52,10 +52,8 @@ public class TemplateChecker {
     private static List<Check> registerChecks(Format format) {
         List<Check> checks = new ArrayList<>();
 
-        Pattern title = format.getTitle();
-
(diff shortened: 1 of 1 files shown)


src/main/java/org/xstefank/check/TemplateChecker.java
```diff
@@ -64,8 +62,14 @@ public void checkPR(JsonNode payload) {
 
         if (format.getCommit() != null) {
             checks.add(new LatestCommitCheck(format.getCommit()));
-        } else if (title != null)
-            checks.add(new LatestCommitCheck(title));
+        }
+
+        Pattern regex = format.getRegex();
```
xstefank: I am sorry, but this is by no means what I had in mind with regards to this issue. Your task is to allow having `title` defined (normal check for title is included) but it is not required that the commit message is the same (matches the same regex) as is defined for title (of course only if the format YAML doesn't contain `commit` tag with the same regex)

Issue #20, opened 2018-08-02
Make error messages for title and commit checks respectively configurable 


xstefank:
Add optional message parameter to `title` and `commit` configuration

Example:
```
format:
  title:
    pattern: ".*"
    message: "Custom error message to be showed"
```

PR #27, opened 2018-09-04
[Issue #20] Make error messages for title and commit checks respectively configurable

khermano:
Issue: #20

### DIFF
--- a/src/main/java/org/xstefank/check/LatestCommitCheck.java
+++ b/src/main/java/org/xstefank/check/LatestCommitCheck.java
@@ -3,18 +3,23 @@ package org.xstefank.check;
 import com.fasterxml.jackson.databind.JsonNode;
 import org.xstefank.api.GitHubAPI;
 import org.xstefank.model.Utils;
-
+import org.xstefank.model.yaml.RegexDefinition;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 public class LatestCommitCheck implements Check {
 
-    private static final String ERROR_MESSAGE = "Invalid commit title format";
+    static final String DEFAULT_MESSAGE = "Invalid commit title content";
 
     private Pattern pattern;
+    private String message;
 
-    public LatestCommitCheck(Pattern pattern) {
-        this.pattern = pattern;
+    public LatestCommitCheck(RegexDefinition commit) {
+        if (commit == null || commit.getPattern() == null) {
+            throw new IllegalArgumentException("Input argument cannot be null!");
+        }
+        this.pattern = commit.getPattern();
+        this.message = (commit.getMessage() != null) ? commit.getMessage() : DEFAULT_MESSAGE;
     }
 
     @Override
@@ -23,7 +28,7 @@ public class LatestCommitCheck implements Check {
         Matcher matcher = pattern.matcher(commitsJson.get(commitsJson.size() - 1).get(Utils.COMMIT).get(Utils.MESSAGE).asText());
 
         if (!matcher.matches()) {
-            return ERROR_MESSAGE;
+            return message;
         }
 
         return null;
--- a/src/main/java/org/xstefank/check/RequiredRowsCheck.java
+++ b/src/main/java/org/xstefank/check/RequiredRowsCheck.java
@@ -2,7 +2,7 @@ package org.xstefank.check;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import org.xstefank.model.Utils;
-import org.xstefank.model.yaml.Row;
+import org.xstefank.model.yaml.RegexDefinition;
 
 import java.util.ArrayList;
 import java.util.Iterator;
@@ -13,22 +13,22 @@ import java.util.regex.Matcher;
 
 public class RequiredRowsCheck implements Check {
 
-    private List<Row> rows;
+    private List<RegexDefinition> rows;
 
-    public RequiredRowsCheck(List<Row> rows) {
+    public RequiredRowsCheck(List<RegexDefinition> rows) {
         this.rows = rows;
     }
 
     @Override
     public String check(JsonNode payload) {
-        List<Row> requiredRows = new ArrayList<>(rows);
+        List<RegexDefinition> requiredRows = new ArrayList<>(rows);
         String description = payload.get(Utils.PULL_REQUEST).get(Utils.BODY).asText();
         Scanner scanner = new Scanner(description);
 
         while (scanner.hasNextLine() && !requiredRows.isEmpty()) {
             String line = scanner.nextLine();
-            for (Iterator<Row> it = requiredRows.iterator(); it.hasNext(); ) {
-                Row row = it.next();
+            for (Iterator<RegexDefinition> it = requiredRows.iterator(); it.hasNext(); ) {
+                RegexDefinition row = it.next();
                 Matcher matcher = row.getPattern().matcher(line);
                 if (matcher.matches()) {
                     requiredRows.remove(row);
@@ -39,7 +39,7 @@ public class RequiredRowsCheck implements Check {
 
         if (!requiredRows.isEmpty()) {
             StringJoiner joiner = new StringJoiner(", ");
-            for (Row row : requiredRows) {
+            for (RegexDefinition row : requiredRows) {
                 joiner.add(row.getMessage());
             }
 
--- a/src/main/java/org/xstefank/check/SkipCheck.java
+++ b/src/main/java/org/xstefank/check/SkipCheck.java
@@ -3,14 +3,15 @@ package org.xstefank.check;
 import com.fasterxml.jackson.databind.JsonNode;
 import org.xstefank.model.Utils;
 import org.xstefank.model.yaml.FormatConfig;
+import org.xstefank.model.yaml.RegexDefinition;
 
 import java.util.regex.Matcher;
 
 public class SkipCheck {
 
     public static boolean shouldSkip(JsonNode payload, FormatConfig config) {
-        if (config == null) {
-            throw new IllegalArgumentException("Argument config is null!");
+        if (payload == null || config == null) {
+            throw new IllegalArgumentException("Input arguments cannot be null!");
         }
         return skipByTitle(payload, config) || skipByCommit(payload, config) || skipByDescriptionFirstRow(payload, config);
     }
@@ -25,7 +26,9 @@ public class SkipCheck {
 
     private static boolean skipByCommit(JsonNode payload, FormatConfig config) {
         if (config.getFormat().getSkipPatterns().getCommit() != null) {
-            LatestCommitCheck latestCommitCheck = new LatestCommitCheck(config.getFormat().getSkipPatterns().getCommit());
+            RegexDefinition commitRegexDefinition = new RegexDefinition();
+            commitRegexDefinition.setPattern(config.getFormat().getSkipPatterns().getCommit());
+            LatestCommitCheck latestCommitCheck = new LatestCommitCheck(commitRegexDefinition);
             return (latestCommitCheck.check(payload) == null);
         }
         return false;
--- a/src/main/java/org/xstefank/check/TemplateChecker.java
+++ b/src/main/java/org/xstefank/check/TemplateChecker.java
@@ -5,7 +5,6 @@ import org.jboss.logging.Logger;
 import org.xstefank.check.additional.AdditionalChecks;
 import org.xstefank.model.yaml.FormatConfig;
 import org.xstefank.model.yaml.Format;
-
 import java.util.ArrayList;
 import java.util.List;
 
@@ -15,19 +14,21 @@ public class TemplateChecker {
     private static final Logger log = Logger.getLogger(TemplateChecker.class);
 
     private List<Check> checks;
-    private FormatConfig config;
 
     public TemplateChecker(FormatConfig config) {
-        if (config == null) {
-            throw new IllegalArgumentException("Argument config cannot be null");
+        if (config == null || config.getFormat() == null) {
+            throw new IllegalArgumentException("Input argument cannot be null!");
         }
-        this.config = config;
         checks = registerChecks(config.getFormat());
     }
 
     public String checkPR(JsonNode payload) {
         log.info("checking PR");
         String description = "";
+
+        if (checks.isEmpty()) {
+            log.warn("No checks were requested in the configuration");
+        }
         for (Check check : checks) {
             String message = check.check(payload);
             if (message != null) {
@@ -54,8 +55,10 @@ public class TemplateChecker {
             checks.add(new LatestCommitCheck(format.getCommit()));
         }
 
-        for (String additional : format.getAdditional()) {
-            checks.add(AdditionalChecks.findCheck(additional));
+        if (format.getAdditional() != null) {
+            for (String additional : format.getAdditional()) {
+                checks.add(AdditionalChecks.findCheck(additional));
+            }
         }
 
         return checks;
--- a/src/main/java/org/xstefank/check/TitleCheck.java
+++ b/src/main/java/org/xstefank/check/TitleCheck.java
@@ -2,25 +2,30 @@ package org.xstefank.check;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import org.xstefank.model.Utils;
-
+import org.xstefank.model.yaml.RegexDefinition;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 public class TitleCheck implements Check {
 
-    private static final String ERROR_MESSAGE = "Invalid title format";
+    static final String DEFAULT_MESSAGE = "Invalid title content";
 
     private Pattern pattern;
+    private String message;
 
-    public TitleCheck(Pattern pattern) {
-        this.pattern = pattern;
+    public TitleCheck(RegexDefinition title) {
+        if (title == null || title.getPattern() == null) {
+            throw new IllegalArgumentException("Input argument cannot be null!");
+        }
+        pattern = title.getPattern();
+        message = (title.getMessage() != null) ? title.getMessage() : DEFAULT_MESSAGE;
     }
 
     @Override
     public String check(JsonNode payload) {
         Matcher matcher = pattern.matcher(payload.get(Utils.PULL_REQUEST).get(Utils.TITLE).asText());
         if (!matcher.matches()) {
-            return ERROR_MESSAGE;
+            return message;
         }
 
         return null;
--- a/src/main/resources/format.yaml
+++ b/src/main/resources/format.yaml
@@ -7,7 +7,12 @@ format:
     title: "((.*)?NO JIRA REQUIRED(.*)\\s?)|(([^do not|don't]\\s)?skip.*template.*check(.*)\\s?)|(([^do not|don't]\\s)?bypass.*template.*check(.*)\\s?)"
     commit: "((.*)?NO JIRA REQUIRED(.*)\\s?)|(([^do not|don't]\\s)?skip.*template.*check(.*)\\s?)|(([^do not|don't]\\s)?bypass.*template.*check(.*)\\s?)"
     description: "((.*)?NO JIRA REQUIRED(.*)\\s?)|(([^do not|don't]\\s)?skip.*template.*check(.*)\\s?)|(([^do not|don't]\\s)?bypass.*template.*check(.*)\\s?)"
-  title: "\\[WFLY-\\d+\\]\\s+.*|WFLY-\\d+\\s+.*"
+  title:
+    pattern: "\\[WFLY-\\d+\\]\\s+.*|WFLY-\\d+\\s+.*"
+    message: "Wrong content of the title!"
+  commit:
+    pattern: "\\[WFLY-\\d+\\]\\s+"
+    message: "Wrong content of the commit message!"
   description:
     required rows:
       - pattern: "JIRA:\\s+https://issues.jboss.org/browse/WFLY-\\d+|https://issues.jboss.org/browse/WFLY-\\d+"
--- a/src/test/java/org/xstefank/TestUtils.java
+++ b/src/test/java/org/xstefank/TestUtils.java
@@ -6,9 +6,10 @@ import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
 import java.io.BufferedReader;
 import java.io.FileReader;
 import java.io.FileWriter;
+import org.xstefank.model.yaml.Format;
 import org.xstefank.model.ConfigTest;
 import org.xstefank.model.yaml.FormatConfig;
-
+import org.xstefank.model.yaml.SkipPatterns;
 import java.io.File;
 import java.io.IOException;
 
@@ -22,7 +23,10 @@ public class TestUtils {
     public static final JsonNode BAD_TEST_PAYLOAD = loadJson(JSON_DIR + "/badTestPayload.json");
     public static final JsonNode ISSUE_PAYLOAD = loadJson(JSON_DIR + "/issuePayload.json");
     public static final JsonNode EMPTY_PAYLOAD = createEmptyJsonPayload();
+    public static final JsonNode TEST_COMMITS_PAYLOAD = loadJson(JSON_DIR + "/testCommitsPayload.json");
 
+    public static final String READ_TOKEN = "readToken";
+    public static final String GET_JSON_WITH_COMMITS = "getJsonWithCommits";
     public static final FormatConfig FORMAT_CONFIG = loadFormatFromYamlFile(YAML_DIR + "/testTemplate.yaml");
     public static final String TEST_CONFIG_PATH = ConfigTest.class.getClassLoader().getResource("testConfig.properties").getPath();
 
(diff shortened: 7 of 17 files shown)


src/test/java/org/xstefank/check/TemplateCheckerTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @BeforeClass
+    public static void beforeClass() throws IOException {
+        File testFile1 = new File(SkipChecker.class.getClassLoader().getResource("testPayload1.json").getFile());
+        testPayload1 = new ObjectMapper().readTree(testFile1);
+    }
+
+    @Test(expected=IllegalArgumentException.class)
+    public void testNullConfigParameterShouldThrowIllegalArgumentException() throws IllegalArgumentException {
+        templateChecker = new TemplateChecker(null);
```
xstefank: nothing wrong, but the assignment is unnecessary

src/main/java/org/xstefank/check/RequiredRowsCheck.java
```diff
@@ -13,34 +12,34 @@
 
 public class RequiredRowsCheck implements Check {
 
-    private List<Row> rows;
+    private List<FormatMatcher> formatMatchers;
```
xstefank: Nope, still this naming doesn't make sense

src/main/java/org/xstefank/check/SkipCheck.java
```diff
@@ -1,41 +1,50 @@
...
+            Matcher titleMatcher = config.getFormat().getSkipPatterns().getTitlePattern().matcher(payload.get(Utils.PULL_REQUEST).get(Utils.TITLE).asText());
             return titleMatcher.matches();
         }
         return false;
     }
 
     private static boolean skipByCommit(JsonNode payload, FormatConfig config) {
-        if (config.getFormat().getSkipPatterns().getCommit() != null) {
-            LatestCommitCheck latestCommitCheck = new LatestCommitCheck(config.getFormat().getSkipPatterns().getCommit());
-            return (latestCommitCheck.check(payload) == null);
```
xstefank: does it make sense to have separate patterns for latestCommitCheck and skipPatterns commit pattern?

src/main/java/org/xstefank/check/TemplateChecker.java
```diff
@@ -28,6 +27,10 @@ public TemplateChecker(FormatConfig config) {
     public String checkPR(JsonNode payload) {
         log.info("checking PR");
         String description = "";
+
+        if (checks.isEmpty()) {
+            log.info("WARNING: No checks were required!");
```
xstefank: :) try checking different methods on the logger

src/main/java/org/xstefank/check/TemplateChecker.java
```diff
@@ -42,20 +45,22 @@ public String checkPR(JsonNode payload) {
     private static List<Check> registerChecks(Format format) {
         List<Check> checks = new ArrayList<>();
 
-        if (format.getTitle() != null) {
-            checks.add(new TitleCheck(format.getTitle()));
+        if (format.getTitleFormatMatcher() != null) {
+            checks.add(new TitleCheck(format.getTitleFormatMatcher()));
```
xstefank: no, this doesn't make sense to me

src/main/resources/format.yaml
```diff
@@ -7,7 +7,12 @@ format:
     title: "((.*)?NO JIRA REQUIRED(.*)\\s?)|(([^do not|don't]\\s)?skip.*template.*check(.*)\\s?)|(([^do not|don't]\\s)?bypass.*template.*check(.*)\\s?)"
     commit: "((.*)?NO JIRA REQUIRED(.*)\\s?)|(([^do not|don't]\\s)?skip.*template.*check(.*)\\s?)|(([^do not|don't]\\s)?bypass.*template.*check(.*)\\s?)"
     description: "((.*)?NO JIRA REQUIRED(.*)\\s?)|(([^do not|don't]\\s)?skip.*template.*check(.*)\\s?)|(([^do not|don't]\\s)?bypass.*template.*check(.*)\\s?)"
-  title: "\\[WFLY-\\d+\\]\\s+.*|WFLY-\\d+\\s+.*"
+  title:
+    pattern: "\\[WFLY-\\d+\\]\\s+.*|WFLY-\\d+\\s+.*"
+    message: "Wrong content of the title!"
+  commit:
+    pattern: "\\[WFLY-\\d+\\]\\s+"
+    message: "Wrong content of the commit!"
```
xstefank: commit message

src/test/java/org/xstefank/check/AuxiliaryTestMethods.java
```diff
@@ -0,0 +1,25 @@
...
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.xstefank.model.yaml.Format;
+import org.xstefank.model.yaml.FormatConfig;
+import org.xstefank.model.yaml.SkipPatterns;
+import java.io.File;
+import java.io.IOException;
+
+public class AuxiliaryTestMethods {
```
xstefank: do not name classes Auxiliary... the classname should clearly state what the class does, but auxiliary can have many meanings, traditionally similar classes are names as utils / utilities

src/test/java/org/xstefank/check/AuxiliaryTestMethods.java
```diff
@@ -0,0 +1,25 @@
...
+
+    public static JsonNode readJsonNode(String resource) throws IOException {
+        File testFile1 = new File(SkipCheck.class.getClassLoader().getResource(resource).getFile());
```
xstefank: why ..1?

src/test/java/org/xstefank/check/AuxiliaryTestMethods.java
```diff
@@ -0,0 +1,25 @@
...
+        return new ObjectMapper().readTree(testFile1);
+    }
+
+    public static FormatConfig helpSetUpSkipPatterns(SkipPatterns testSkipPatterns) {
```
xstefank: this method returns configured format configuration with skipPatterns if I understand correctly. It doesn't help with anything :)

src/test/java/org/xstefank/check/LatestCommitCheckTest.java
```diff
@@ -0,0 +1,80 @@
...
+    @Test
+    public void testCheckSimpleRegexMatch() {
+        testCommitFormatMatcher.setPattern(Pattern.compile("test commit"));
+        testLatestCommitCheck = new LatestCommitCheck(testCommitFormatMatcher);
+
+        Assert.assertNull("Cannot match valid regex", testLatestCommitCheck.check(testPullRequestPayload));
+    }
+
+    @Test
+    public void testCheckSimpleRegexNonMatchReturnExpectedMessage() {
```
xstefank: Returns

src/test/java/org/xstefank/check/LatestCommitCheckTest.java
```diff
@@ -0,0 +1,80 @@
...
+        testCommitFormatMatcher.setPattern(Pattern.compile("can't.*match.*this"));
+        testCommitFormatMatcher.setMessage("This is testCommitFormatMatcher message.");
+        testLatestCommitCheck = new LatestCommitCheck(testCommitFormatMatcher);
+
+        Assert.assertNotNull("Matched invalid regex", testLatestCommitCheck.check(testPullRequestPayload));
+        Assert.assertEquals("Not returned expected message", "This is testCommitFormatMatcher message.", testLatestCommitCheck.check(testPullRequestPayload));
```
xstefank: this should be a full sentence, it won't make sense to the caller when this test fails - you can write "Unexpected message returned" or "method call did not return..."

src/test/java/org/xstefank/check/SkipCheckTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+@RunWith(PowerMockRunner.class)
+@PrepareForTest(GitHubAPI.class)
+public class SkipCheckTest {
+
+    private static JsonNode testPullRequestPayload;
+    private static JsonNode testCommitsPayload;
+    private static FormatConfig testFormatConfig = new FormatConfig();
+    private SkipPatterns testSkipPatterns = new SkipPatterns();
+    private static final String stubMethodName = "getJsonWithCommits";
```
xstefank: there are still at least 2 places in these tests that will need to be changed in case this method is renamed

src/test/java/org/xstefank/check/SkipCheckTest.java
```diff
@@ -0,0 +1,114 @@
...
+    public void testNullPayloadParameter() throws IllegalArgumentException{
+        SkipCheck.shouldSkip(null, testFormatConfig);
+    }
+
+    @Test
+    public void testSkipByTitleRegexMatch(){
+        testSkipPatterns.setTitlePattern(Pattern.compile("test title"));
+        testFormatConfig = helpSetUpSkipPatterns(testSkipPatterns);
+
+        Assert.assertTrue("Method shouldSkip should by inner method skipByTitle returned true", SkipCheck.shouldSkip(testPullRequestPayload, testFormatConfig));
```
xstefank: please rephrase this

src/test/java/org/xstefank/check/SkipCheckTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+        Assert.assertTrue("Method shouldSkip should by inner method skipByPullRequestDescriptionFirstRow returned true",SkipCheck.shouldSkip(testPullRequestPayload, testFormatConfig));
+    }
+
+    @Test
+    public void testSkipByPullRequestDescriptionRegexNonMatch(){
+        testSkipPatterns.setDescriptionPattern(Pattern.compile("can't.*match.*this"));
+        testFormatConfig = helpSetUpSkipPatterns(testSkipPatterns);
+
+        Assert.assertFalse("Method shouldSkip should by inner method skipByPullRequestDescriptionFirstRow returned false",SkipCheck.shouldSkip(testPullRequestPayload, testFormatConfig));
```
xstefank: skipByPullRequestDescriptionFirstRow?

src/test/java/org/xstefank/check/SkipCheckTest.java
```diff
@@ -0,0 +1,122 @@
...
+    private static FormatConfig testFormatConfig = new FormatConfig();
+    private SkipPatterns testSkipPatterns = new SkipPatterns();
+    private RegexDefinition testTitle =  new RegexDefinition();
+    private RegexDefinition testCommit =  new RegexDefinition();
+    private RegexDefinition testDescription =  new RegexDefinition();
+
+    @BeforeClass
+    public static void beforeClass() throws IOException {
+        testPullRequestPayload = readJsonNode(TestUtilities.getTestPullRequestPayloadName());
+        testCommitsPayload = readJsonNode(TestUtilities.getTestCommitsPayloadName());
```
xstefank: same setup is done in the previous class too, this is not precisely an error, but it would be better to have setup only in one place -- utility or shared ancestor

src/test/java/org/xstefank/check/SkipCheckTest.java
```diff
@@ -0,0 +1,122 @@
...
+        SkipCheck.shouldSkip(null, testFormatConfig);
+    }
+
+    @Test
+    public void testSkipByTitleRegexMatch(){
+        testTitle.setPattern(Pattern.compile("test title"));
+        testSkipPatterns.setTitle(testTitle);
+        testFormatConfig = setUpFormatConfig(testSkipPatterns);
+
+        Assert.assertTrue("Method skipByTitle cannot match valid regex", SkipCheck.shouldSkip(testPullRequestPayload, testFormatConfig));
```
xstefank: again "don't forget to update error message when you change the private method somehow"

src/test/java/org/xstefank/check/SkipCheckTest.java
```diff
@@ -0,0 +1,122 @@
...
+    public void testSkipByPullRequestDescriptionRegexNonMatch(){
+        testDescription.setPattern(Pattern.compile("can't.*match.*this"));
+        testSkipPatterns.setDescription(testDescription);
+        testFormatConfig = setUpFormatConfig(testSkipPatterns);
+
+        Assert.assertFalse("Method skipByPullRequestDescription matched invalid regex",SkipCheck.shouldSkip(testPullRequestPayload, testFormatConfig));
+    }
+
+    @Test
+    public void testShouldSkipNullSkipPatterns() {
```
xstefank: you are not passing null anywhere in this method, default or empty would be better

src/test/java/org/xstefank/check/TestUtilities.java
```diff
@@ -0,0 +1,45 @@
...
+
+    public static FormatConfig setUpFormatConfig(SkipPatterns testSkipPatterns) {
+        Format testFormat = new Format();
+        testFormat.setSkipPatterns(testSkipPatterns);
+        FormatConfig testFormatConfig = new FormatConfig();
+        testFormatConfig.setFormat(testFormat);
+        return testFormatConfig;
+    }
+
+    public static String getStubMethodName() {
```
xstefank: sorry but this is wrong, will StubMethod always be getJsonWithCommits? why to name is as stub method in utils class, when I as a user call `TestUtilities.getStubMethodName` I wouldn't by any chance expect the one exact string returned by this method

src/test/java/org/xstefank/check/TestUtilities.java
```diff
@@ -0,0 +1,45 @@
...
+        return "getJsonWithCommits";
```
xstefank: also a static returning only a constant string is redundant, this should be stored in public static final field as a constant

src/test/java/org/xstefank/check/TitleCheckTest.java
```diff
@@ -1,37 +1,61 @@
...
 
+import static org.xstefank.check.TestUtilities.readJsonNode;
+
 public class TitleCheckTest {
 
-    private static JsonNode testPayload;
-    private TitleCheck titleCheck;
+    private static JsonNode testSimplePullRequestPayload;
+    private RegexDefinition testTitle = new RegexDefinition();
+    private TitleCheck testTitleCheck;
```
xstefank: why this renaming?

src/test/java/org/xstefank/check/TestUtilities.java
```diff
@@ -0,0 +1,64 @@
...
+    public static FormatConfig setUpFormatConfig(SkipPatterns testSkipPatterns) {
+        Format testFormat = new Format();
+        testFormat.setSkipPatterns(testSkipPatterns);
+        FormatConfig testFormatConfig = new FormatConfig();
+        testFormatConfig.setFormat(testFormat);
+        return testFormatConfig;
+    }
+
+    public static void stubMethodWithPowerMock(JsonNode testCommitsPayload) {
+        suppress(method(GitHubAPI.class, SUPPRESS_METHOD_NAME));
```
xstefank: you can name this constant with reference of the name of the suppressed method, in that case, when we will need to suppress another method there is no need to rename also this one

src/test/java/org/xstefank/check/TestUtilities.java
```diff
@@ -0,0 +1,64 @@
...
+        PowerMockito.stub(PowerMockito.method(GitHubAPI.class, STUB_METHOD_NAME, JsonNode.class)).toReturn(testCommitsPayload);
```
xstefank: see above

src/test/java/org/xstefank/check/TestUtilities.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    public static JsonNode getTestPullRequestPayload() throws IOException {
```
xstefank: are these 3 methods called only on one place? in that case they are unnecessary

src/test/java/org/xstefank/check/TestUtilities.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    public static JsonNode getTestCommitsPayload() throws IOException {
+        return readJsonNode(TEST_COMMITS_PAYLOAD);
+    }
+
+    public static JsonNode getTestSimplePullRequestPayload() throws IOException {
+        return readJsonNode(TEST_SIMPLE_PULL_REQUEST_PAYLOAD);
+    }
+}
```
xstefank: there was also a similar test utility class merged some time ago, so take care that no code is unnecessary duplicated please

xstefank:
there are conflicts with the current master, please rebase this PR

src/test/java/org/xstefank/TestUtils.java
```diff
@@ -45,4 +54,17 @@ private static JsonNode createEmptyJsonPayload() {
...
\ No newline at end of file
+    public static FormatConfig setUpFormatConfig(SkipPatterns testSkipPatterns) {
+        Format testFormat = new Format();
+        testFormat.setSkipPatterns(testSkipPatterns);
+        FormatConfig testFormatConfig = new FormatConfig();
+        testFormatConfig.setFormat(testFormat);
+        return testFormatConfig;
+    }
+
+    public static void stubMethodWithPowerMock() {
```
xstefank: for a method with a such general name I would expect that it will take an argument stating the name of the stubbed method or that the name would directly incline what is happening inside, when I call this in user code as `TestUtils.stubMethodWithPowerMock()` I cannot possibly know what is happening without checking the content of this method

src/test/java/org/xstefank/TestUtils.java
```diff
@@ -45,4 +54,17 @@ private static JsonNode createEmptyJsonPayload() {
...
+        suppress(method(GitHubAPI.class, READ_TOKEN));
```
xstefank: and it is doing actually a little bit more than just stubbing

src/test/java/org/xstefank/check/LatestCommitCheckTest.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @Test
+    public void testCheckSimpleRegexNonMatchReturnsExpectedMessage() {
+        testCommit.setPattern(Pattern.compile("can't.*match.*this"));
+        testCommit.setMessage("This is testCommit message.");
+        latestCommitCheck = new LatestCommitCheck(testCommit);
+
+        Assert.assertNotNull("Matched invalid regex", latestCommitCheck.check(TestUtils.TEST_PAYLOAD));
+        Assert.assertEquals("Unexpected message returned", "This is testCommit message.", latestCommitCheck.check(TestUtils.TEST_PAYLOAD));
```
xstefank: if you store the result, you don't need to call this method with the same argument two times

src/test/java/org/xstefank/check/SkipCheckTest.java
```diff
@@ -0,0 +1,91 @@
...
+import org.xstefank.model.yaml.SkipPatterns;
+import java.util.regex.Pattern;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
+
+@RunWith(PowerMockRunner.class)
+@PrepareForTest(GitHubAPI.class)
+public class SkipCheckTest {
+
+    private static FormatConfig testFormatConfig = new FormatConfig();
```
xstefank: the instance you create here is actually overriden in every method that uses this field by the call to `TestUtils.setUpFormatConfig` so it's not necessary to be created

src/test/java/org/xstefank/check/SkipCheckTest.java
```diff
@@ -0,0 +1,91 @@
...
+        Assert.assertFalse("Method matched invalid description regex",SkipCheck.shouldSkip(TestUtils.TEST_PAYLOAD, testFormatConfig));
+    }
+
+    @Test
+    public void testShouldSkipEmptySkipPatterns() {
+        TestUtils.stubMethodWithPowerMock();
+
+        testFormatConfig = TestUtils.setUpFormatConfig(testSkipPatterns);
+
+        Assert.assertFalse("Method works without defining regex",SkipCheck.shouldSkip(TestUtils.TEST_PAYLOAD, testFormatConfig));
```
xstefank: to the method is supposed to define a regex somewhere?

src/test/java/org/xstefank/check/LatestCommitCheckTest.java
```diff
@@ -0,0 +1,67 @@
...
+        Assert.assertEquals("Unexpected message returned", "This is testCommit message.", result);
+    }
+
+    @Test
+    public void testCheckSimpleRegexNonMatchReturnsDefaultMessage() {
+        testCommit.setPattern(Pattern.compile("can't.*match.*this"));
+        latestCommitCheck = new LatestCommitCheck(testCommit);
+
+        Assert.assertEquals("Unexpected message returned", LatestCommitCheck.DEFAULT_MESSAGE, latestCommitCheck.check(TestUtils.TEST_PAYLOAD));
+    }
```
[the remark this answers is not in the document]
xstefank: actually interesting read: https://www.baeldung.com/junit-5-test-order

src/main/java/org/xstefank/check/LatestCommitCheck.java
```diff
@@ -3,18 +3,24 @@
...
 
     private Pattern pattern;
+    private String message;
 
-    public LatestCommitCheck(Pattern pattern) {
-        this.pattern = pattern;
+    public LatestCommitCheck(RegexDefinition commit) {
+        if (commit == null || commit.getPattern() == null) {
+            throw new IllegalArgumentException("Input argument cannot be null!");
+        } else {
```
xstefank: this else block is not necessary

src/main/java/org/xstefank/check/SkipCheck.java
```diff
@@ -25,8 +26,9 @@ private static boolean skipByTitle(JsonNode payload, FormatConfig config) {
 
     private static boolean skipByCommit(JsonNode payload, FormatConfig config) {
         if (config.getFormat().getSkipPatterns().getCommit() != null) {
-            LatestCommitCheck latestCommitCheck = new LatestCommitCheck(config.getFormat().getSkipPatterns().getCommit());
-            return (latestCommitCheck.check(payload) == null);
+            JsonNode commitsJson = GitHubAPI.getJsonWithCommits(payload);
+            Matcher commitMatcher = config.getFormat().getSkipPatterns().getCommit().matcher(commitsJson.get(commitsJson.size() - 1).get(Utils.COMMIT).get(Utils.MESSAGE).asText());
```
xstefank: there is no particular issue with directly calling what I assume LatestCommitCheck is doing. However if commit JSON structure changes (by github, we don't have influence on such things) on have many places will we need to hunt down such change?

src/test/java/org/xstefank/check/LatestCommitCheckTest.java
```diff
@@ -0,0 +1,68 @@
...
+import org.xstefank.api.GitHubAPI;
+import org.xstefank.model.yaml.RegexDefinition;
+import java.util.regex.Pattern;
+import static org.powermock.api.support.membermodification.MemberMatcher.method;
+
+@RunWith(PowerMockRunner.class)
+@PrepareForTest(GitHubAPI.class)
+public class LatestCommitCheckTest {
+
+    private RegexDefinition testCommit;
```
xstefank: maybe please rename this

src/test/java/org/xstefank/check/TitleCheckTest.java
```diff
@@ -3,22 +3,48 @@
 import org.junit.Assert;
 import org.junit.Test;
 import org.xstefank.TestUtils;
-
+import org.xstefank.model.yaml.RegexDefinition;
 import java.util.regex.Pattern;
 
 public class TitleCheckTest {
 
+    private RegexDefinition testTitle = new RegexDefinition();
```
xstefank: also please rename - just to clearly see that it's not directly title but more title validation definition

src/main/java/org/xstefank/check/TitleCheck.java
```diff
@@ -2,27 +2,33 @@
...
 
     private Pattern pattern;
+    private String message;
 
-    public TitleCheck(Pattern pattern) {
-        this.pattern = pattern;
+    public TitleCheck(RegexDefinition title) {
+        if (title == null || title.getPattern() == null) {
+            throw new IllegalArgumentException("Input argument cannot be null!");
+        } else {
```
xstefank: this else block is also unnecessary ;) please try to always check comments also for other occurrences

src/test/java/org/xstefank/check/TitleCheckTest.java
```diff
@@ -3,22 +3,48 @@
...
+        titleCheck = new TitleCheck(titleRegexDefinition);
+
         Assert.assertNotNull("Matched invalid regex", titleCheck.check(TestUtils.TEST_PAYLOAD));
+        Assert.assertEquals("Unexpected message returned", titleRegexDefinition.getMessage(), titleCheck.check(TestUtils.TEST_PAYLOAD));
+    }
+
+    @Test
+    public void testCheckSimpleRegexNonMatchReturnsDefaultMessage() {
+        titleRegexDefinition.setPattern(Pattern.compile("can't.*match.*this"));
+        titleCheck = new TitleCheck(titleRegexDefinition);
```
xstefank: you do not clean titleRegexDefinition instance. if `testCheckSimpleRegexNonMatchReturnsExpectedMessage` runs just before `testCheckSimpleRegexNonMatchReturnsDefaultMessage` will the message be set to DEFAULT_MESSAGE?

PR #29, opened 2018-09-10
[Issue #23] Add possibility to load format configuration from URL

lukasburda:
Issue: https://github.com/xstefank/tyr/issues/23


src/main/java/org/xstefank/webhook/WebHookEndpoint.java
```diff
@@ -44,14 +47,30 @@ public void processPullRequest(JsonNode payload) {
...
+            try {
+                inputStream = new URL(url).openStream();
+                byte[] buffer = new byte[inputStream.available()];
+                inputStream.read(buffer);
+                outputStream = new FileOutputStream(configFile);
+                outputStream.write(buffer);
+            } catch (IOException e) {
+                throw new IllegalArgumentException("Cannot load configuration file", e);
+            } finally {
+                try {
```
xstefank: why inner try block?

src/main/java/org/xstefank/webhook/WebHookEndpoint.java
```diff
@@ -44,14 +47,30 @@ public void processPullRequest(JsonNode payload) {
     }
 
     private static FormatConfig readConfig() {
-        String configFileName = System.getProperty(TEMPLATE_FORMAT_FILE);
-        if (configFileName == null) {
-            configFileName = System.getProperty(Utils.JBOSS_CONFIG_DIR) + "/format.yaml";
```
xstefank: so now you are loading format only from URL and and format.yaml in configuration. Please reverse the original property option

src/main/java/org/xstefank/webhook/WebHookEndpoint.java
```diff
@@ -44,14 +50,27 @@ public void processPullRequest(JsonNode payload) {
...
         if (configFileName == null) {
             configFileName = System.getProperty(Utils.JBOSS_CONFIG_DIR) + "/format.yaml";
         }
         log.info(configFileName);
         File configFile = new File(configFileName);
-        ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
 
+        if (url != null && !configFile.exists()) {
+            try (InputStream inputStream = new URL(url).openStream();
+                 OutputStream outputStream = new FileOutputStream(configFile)) {
```
xstefank: this is error - it will override format.yaml in standalone config dir or passed by property

src/main/java/org/xstefank/webhook/WebHookEndpoint.java
```diff
@@ -44,14 +50,27 @@ public void processPullRequest(JsonNode payload) {
...
+                byte[] buffer = new byte[inputStream.available()];
```
xstefank: this is wrong, check the second paragraph in the method javadoc... https://docs.oracle.com/javase/7/docs/api/java/io/InputStream.html#available()

src/main/java/org/xstefank/webhook/WebHookEndpoint.java
```diff
@@ -44,14 +50,27 @@ public void processPullRequest(JsonNode payload) {
     }
 
     private static FormatConfig readConfig() {
+        String url = Utils.readTemplateConfigUrlFromProperties(System
+                .getProperty(Utils.JBOSS_CONFIG_DIR), Utils.CONFIG_FILE);
```
xstefank: also it would be appreciated to allow users to pass the URL through system propery

src/main/java/org/xstefank/webhook/WebHookEndpoint.java
```diff
@@ -44,16 +47,21 @@ public void processPullRequest(JsonNode payload) {
...
 
     private static FormatConfig readConfig() {
-        String configFileName = System.getProperty(TEMPLATE_FORMAT_FILE);
-        if (configFileName == null) {
-            configFileName = System.getProperty(Utils.JBOSS_CONFIG_DIR) + "/format.yaml";
-        }
-        log.info(configFileName);
-        File configFile = new File(configFileName);
+        String target = Utils.readTemplateConfigUrlFromProperties(System
+                .getProperty(Utils.JBOSS_CONFIG_DIR), Utils.CONFIG_FILE);
```
xstefank: please add an option to set URL on command line

src/main/java/org/xstefank/model/Utils.java
```diff
@@ -28,6 +29,14 @@
...
 
     public static String readTokenFromProperties(String dirName, String fileName) {
+        return readPropertyFromFile(TOKEN_PROPERTY, dirName, fileName);
+    }
+
+    public static String readTemplateConfigUrlFromProperties(String dirName, String fileName) {
+        return readPropertyFromFile(TEMPLATE_FORMAT_URL, dirName, fileName);
+    }
+
+    private static String readPropertyFromFile(String property, String dirName, String fileName) {
```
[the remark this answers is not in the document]
xstefank: i don't understand how you want to apply singleton in this case? just make sure you won't load properties file individually for each property (you always load whole properties file when this method is called)

src/main/java/org/xstefank/model/PropertiesLoader.java
```diff
@@ -0,0 +1,67 @@
...
+
+public class PropertiesLoader {
+
+    private static PropertiesLoader propertiesLoader = null;
+    private Properties properties;
+
+    private PropertiesLoader() {
+        String dirProperty = System.getProperty(JBOSS_CONFIG_DIR);
+        if (dirProperty != null) {
+            File dir = new File(System.getProperty(JBOSS_CONFIG_DIR));
```
xstefank: why do you call `System.getProperty` again? you have the result stored in `dirProperty`

src/main/java/org/xstefank/model/PropertiesLoader.java
```diff
@@ -0,0 +1,67 @@
...
+        if (propertiesLoader == null) {
+            propertiesLoader = new PropertiesLoader();
+        }
+        return propertiesLoader;
+    }
+
+    public Properties getProperties() {
+        if (properties != null)
+            return properties;
+        throw new NullPointerException("Properties were not load properly");
```
xstefank: this is not a way of how to handle errors, you should fail the application when you know the properties haven't been loaded properly, not when somebody is accessing them, this would result into a stactrace that has nothing to do with the actual error

src/main/java/org/xstefank/model/PropertiesLoader.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    private Properties readPropertiesFromFile(File targetFile) {
+        InputStream is = null;
+        try {
+            is = new FileInputStream(targetFile);
+            Properties properties = new Properties();
+            properties.load(is);
+
+            if (properties.size() > 0) {
```
xstefank: so if somebody passes empty properties file the application should later fail with NPE?

src/main/java/org/xstefank/model/PropertiesLoader.java
```diff
@@ -0,0 +1,67 @@
...
+                return properties;
+            }
+        } catch (Exception e) {
+            // intentionally ignored
```
xstefank: no, this is wrong, here you have the actual error and you are just throwing it away...

src/main/java/org/xstefank/model/PropertiesLoader.java
```diff
@@ -0,0 +1,67 @@
...
+        } finally {
+            try {
+                if (is != null) {
+                    is.close();
+                }
+            } catch (IOException e) {
+                // intentionally ignored
```
xstefank: please describe your intention here

src/main/java/org/xstefank/model/PropertiesLoader.java
xstefank: or you can describe your intention

src/main/java/org/xstefank/model/Utils.java
```diff
@@ -27,31 +22,13 @@
...
-    public static String readTokenFromProperties(String dirName, String fileName) {
-        InputStream is = null;
-        File dir = new File(dirName);
-        File fileProp = new File(dir, fileName);
-
-        try {
-            is = new FileInputStream(fileProp);
-            Properties properties = new Properties();
-            properties.load(is);
+    private static PropertiesLoader propertiesLoader = PropertiesLoader.getInstance();
```
xstefank: I have nothing against making `PropertiesLoader` a singleton, but why is it necessary? you are using it only in this one place

xstefank: also I don't think there is a necessity for yet another class that just loads the properties, make it configuration loader that handles all of the configuration loading processing, or leave it here in `Utils` and just store the loaded properties in a field for further use; if you think the singleton wrapper you've implemented is better solution, please comment here your reasoning

src/main/java/org/xstefank/model/Utils.java
```diff
@@ -27,31 +22,13 @@
...
-            try {
-                if (is != null) {
-                    is.close();
-                }
-            } catch (IOException e) {
-                // intentionally ignored
-            }
-        }
+    public static String getTokenFromProperties() {
+        return propertiesLoader.getProperties().getProperty(TOKEN_PROPERTY);
```
xstefank: this seems to me really forced and unnecessary

src/main/java/org/xstefank/webhook/WebHookEndpoint.java
```diff
@@ -44,18 +47,27 @@ public void processPullRequest(JsonNode payload) {
...
+            String configFileName = getTemplateFileName();
+            log.info(configFileName);
+            File configFile = new File(configFileName);
             return mapper.readValue(configFile, FormatConfig.class);
         } catch (IOException e) {
             throw new IllegalArgumentException("Cannot load configuration file", e);
         }
     }
+
+    private static String getUrl() {
```
xstefank: this can directly return `URL`

src/main/java/org/xstefank/webhook/WebHookEndpoint.java
```diff
@@ -44,18 +47,27 @@ public void processPullRequest(JsonNode payload) {
...
+        String url = System.getProperty(TEMPLATE_FORMAT_URL);
+        return url == null ? Utils.getTemplateUrlFromProperties() : url;
+    }
+
+    private static String getTemplateFileName() {
```
xstefank: if you chose to include url in the properties configuration too, you should also align potential path in case it's not defined by the system property

src/test/java/org/xstefank/YamlConfigTest.java
```diff
@@ -16,7 +17,9 @@ public void testValidTemplateConfig() {
     @Test
     public void readTokenTest() {
         String pathToTestConfig = YamlConfigTest.class.getClassLoader().getResource("testConfig.properties").getPath();
+        PropertiesLoader propertiesLoader = PropertiesLoader.getInstance();
+        propertiesLoader.savePropertiesFromFile(pathToTestConfig);
```
xstefank: `PropertiesLoader` is obviously defined by the loaded properties file, such dependencies are typically passed by the constructor. Please explain why the singleton is needed here

src/main/java/org/xstefank/model/Utils.java
```diff
@@ -27,31 +29,30 @@
...
+
+    public static Properties getProperties(String dirName, String fileName) {
+        if (properties == null) {
+            File dir = new File(dirName);
+            File fileProp = new File(dir, fileName);
+            try (InputStream is = new FileInputStream(fileProp)) {
+                Properties properties = new Properties();
+                properties.load(is);
+                if (properties.isEmpty())
+                    throw new RuntimeException("Properties file cannot be empty");
```
xstefank: why this? I don't think the application needs to be (probably) terminated because of the empty properties file

src/main/java/org/xstefank/model/Utils.java
```diff
@@ -27,31 +29,30 @@
...
+                Utils.properties = properties;
```
xstefank: no need to prefix this with `Utils.` as you are in the class containing this field

src/main/java/org/xstefank/api/GitHubAPI.java
```diff
@@ -73,7 +75,7 @@ private static URI getUriFromPayload(JsonNode payload) {
     }
 
     private static String readToken() {
-        String token = readTokenFromProperties(System.getProperty(Utils.JBOSS_CONFIG_DIR), Utils.CONFIG_FILE);
+        String token = Utils.getProperties(System.getProperty(JBOSS_CONFIG_DIR), CONFIG_FILE).getProperty(TOKEN_PROPERTY);
```
xstefank: you are still loading same properties file several times --- 1/

src/main/java/org/xstefank/model/Utils.java
```diff
@@ -27,31 +29,30 @@
...
             }
         }
+        return properties;
+    }
 
-        return null;
+    public static URL getFormatFileUrl() throws MalformedURLException {
+        String target = System.getProperty(TEMPLATE_FORMAT_URL);
+        if (target == null)
+            target = getProperties(System.getProperty(JBOSS_CONFIG_DIR), CONFIG_FILE).getProperty(TEMPLATE_FORMAT_URL);
```
xstefank: you are still loading same properties file several times --- 2/

src/test/java/org/xstefank/api/GitHubAPITest.java
```diff
@@ -6,18 +6,16 @@
...
-
 @RunWith(PowerMockRunner.class)
 @PrepareForTest({GitHubAPI.class, Utils.class})
 public class GitHubAPITest {
 
     @Before
     public void before() throws Exception {
-        PowerMockito.mockStatic(Utils.class);
-        PowerMockito.when(Utils.readTokenFromProperties(anyString(), anyString())).thenReturn(null);
+        Utils.getProperties("", TestUtils.PROPERTIES_FILE_PATH);
```
xstefank: why are you reading this if you are not using it anywhere?

Issue #26, opened 2018-09-04
Implement control for the repository format in format.yaml


lukasburda:

xstefank:
Are you thinking about traditional check implementing the `Check` interface? In that case where the error would be displayed?

lukasburda:
I think it would be better to implement this feature to function where format.yaml loads and if there is bad format throw some exception.

xstefank:
@lukasburda so the current implementation loads the application when invalid format.yaml is passed in?

lukasburda:
@xstefank yes it does

lukasburda:
No, but if you type to format.yaml bad repository value like "user/repo/somethingelse" or "userrepo" it loads. I think it would be better to warn user that it has bad format.

xstefank:
I understand now your concern, please go ahead, looking forward to the PR!

PR #30, opened 2018-09-10
[Issue #26] Implement control for the repository format in format.yaml

lukasburda:
Issue: https://github.com/xstefank/tyr/issues/26

### DIFF
--- a/src/main/java/org/xstefank/verification/RepositoryFormatVerification.java
+++ b/src/main/java/org/xstefank/verification/RepositoryFormatVerification.java
@@ -1,0 +1,12 @@
+package org.xstefank.verification;
+
+import org.xstefank.model.yaml.FormatConfig;
+
+public class RepositoryFormatVerification implements Verification {
+
+    @Override
+    public void verify(FormatConfig formatConfig) throws InvalidConfigurationException {
+        if (!formatConfig.getRepository().matches("^[a-zA-Z0-9_]*/[a-zA-Z0-9_-]*$"))
+            throw new InvalidConfigurationException("Wrong repository format in configuration file");
+    }
+}
--- a/src/main/java/org/xstefank/verification/Verification.java
+++ b/src/main/java/org/xstefank/verification/Verification.java
@@ -1,0 +1,8 @@
+package org.xstefank.verification;
+
+import org.xstefank.model.yaml.FormatConfig;
+
+public interface Verification {
+
+    void verify(FormatConfig formatConfig) throws InvalidConfigurationException;
+}
--- a/src/main/java/org/xstefank/verification/VerificationHandler.java
+++ b/src/main/java/org/xstefank/verification/VerificationHandler.java
@@ -1,0 +1,26 @@
+package org.xstefank.verification;
+
+import org.xstefank.model.yaml.FormatConfig;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class VerificationHandler {
+
+    private static final List<Verification> verifications = registerVerifications();
+
+    public static void verifyConfiguration(FormatConfig formatConfig) throws InvalidConfigurationException {
+        for (Verification verification : verifications) {
+            verification.verify(formatConfig);
+        }
+    }
+
+    private static List<Verification> registerVerifications() {
+        List<Verification> verifications = new ArrayList<>();
+
+        verifications.add(new RepositoryFormatVerification());
+        verifications.add(new FormatElementVerification());
+
+        return verifications;
+    }
+}
--- a/src/main/java/org/xstefank/webhook/WebHookEndpoint.java
+++ b/src/main/java/org/xstefank/webhook/WebHookEndpoint.java
@@ -7,15 +7,19 @@ import org.jboss.logging.Logger;
 import org.xstefank.api.GitHubAPI;
 import org.xstefank.check.SkipCheck;
 import org.xstefank.check.TemplateChecker;
+import org.xstefank.verification.InvalidConfigurationException;
+import org.xstefank.verification.VerificationHandler;
 import org.xstefank.model.CommitStatus;
 import org.xstefank.model.Utils;
 import org.xstefank.model.yaml.FormatConfig;
+
 import javax.ws.rs.Consumes;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.core.MediaType;
 import java.io.File;
 import java.io.IOException;
+
 import static org.xstefank.check.TemplateChecker.TEMPLATE_FORMAT_FILE;
 
 @Path("/")
@@ -51,10 +55,11 @@ public class WebHookEndpoint {
         log.info(configFileName);
         File configFile = new File(configFileName);
         ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
-
         try {
-            return mapper.readValue(configFile, FormatConfig.class);
-        } catch (IOException e) {
+            FormatConfig formatConfig = mapper.readValue(configFile, FormatConfig.class);
+            VerificationHandler.verifyConfiguration(formatConfig);
+            return formatConfig;
+        } catch (IOException | InvalidConfigurationException e) {
             throw new IllegalArgumentException("Cannot load configuration file", e);
         }
     }
--- a/src/test/java/org/xstefank/verification/VerificationsTest.java
+++ b/src/test/java/org/xstefank/verification/VerificationsTest.java
@@ -1,0 +1,34 @@
+package org.xstefank.verification;
+
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.xstefank.model.yaml.FormatConfig;
+
+import static org.xstefank.TestUtils.loadFormatFromYamlFile;
+import static org.xstefank.TestUtils.YAML_DIR;
+import static org.xstefank.TestUtils.FOMAT_CONFIG;
+
+public class VerificationsTest {
+
+    private static FormatConfig badFormatConfig;
+
+    @BeforeClass
+    public static void beforeClass() {
+        badFormatConfig = loadFormatFromYamlFile(YAML_DIR + "/wrongTemplate.yaml");
+    }
+
+    @Test
+    public void testReadValidFormatConfiguration() throws InvalidConfigurationException {
+        VerificationHandler.verifyConfiguration(FOMAT_CONFIG);
+    }
+
+    @Test(expected = InvalidConfigurationException.class)
+    public void testReadInvalidFormatConfiguration() throws InvalidConfigurationException {
+        VerificationHandler.verifyConfiguration(badFormatConfig);
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void testNullParameterToVerify() throws InvalidConfigurationException {
+        VerificationHandler.verifyConfiguration(null);
+    }
+}
--- a/src/main/java/org/xstefank/verification/FormatElementVerification.java
+++ b/src/main/java/org/xstefank/verification/FormatElementVerification.java
@@ -1,0 +1,12 @@
+package org.xstefank.verification;
+
+import org.xstefank.model.yaml.FormatConfig;
+
+public class FormatElementVerification implements Verification {
+
+    @Override
+    public void verify(FormatConfig formatConfig) throws InvalidConfigurationException {
+        if (formatConfig.getFormat() == null)
+            throw new InvalidConfigurationException("Element 'format' is not specified");
+    }
+}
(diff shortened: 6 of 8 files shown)


src/main/java/org/xstefank/webhook/WebHookEndpoint.java
```diff
@@ -53,9 +55,18 @@ private static FormatConfig readConfig() {
...
 
         try {
-            return mapper.readValue(configFile, FormatConfig.class);
+            FormatConfig formatConfig = mapper.readValue(configFile, FormatConfig.class);
+            String repoFormat = formatConfig.getRepository();
+            if (repoFormat.matches("^[A-z,0-9].*/[A-z,0-9,-].*$") && repoFormat.split("/").length == 2) {
+                return formatConfig;
+            }
+            else
+                throw new BadRepositoryFormatException("Bad repository format in configuration file");
```
xstefank: do you think this check should be here? what if more checks will be added in the future. It would be useful to define some kind of external validation

src/main/java/org/xstefank/control/Control.java
```diff
@@ -0,0 +1,8 @@
+package org.xstefank.control;
+
+import org.xstefank.model.yaml.FormatConfig;
+
+public interface Control {
```
xstefank: can you rename this to Verification please

src/main/java/org/xstefank/control/ControlHandler.java
```diff
@@ -0,0 +1,40 @@
+package org.xstefank.control;
+
+import org.xstefank.model.yaml.FormatConfig;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ControlHandler {
```
xstefank: all control rename to verification, control is rather restrictive

src/main/java/org/xstefank/control/ControlHandler.java
```diff
@@ -0,0 +1,40 @@
...
+            String message = null;
+            for (Control control : controls) {
+                message = control.control(formatConfig);
+            }
+            if (message == null) {
+                return true;
+            } else {
+                throw new BadFormatException(message);
+            }
+        } catch (BadFormatException e) {
```
xstefank: you are catching exception you are throwing in try block

src/main/java/org/xstefank/control/RepositoryFormatControl.java
```diff
@@ -0,0 +1,18 @@
...
+
+public class RepositoryFormatControl implements Control {
+
+    private final String REGEX = "^[A-z,0-9].*/[A-z,0-9,-].*$";
+    private final String MESSAGE = "Bad repository format in configuration file";
+
+    public String control(FormatConfig formatConfig) {
+        String repoFormat = formatConfig.getRepository();
+        if (repoFormat.matches(REGEX) && repoFormat.split("/").length == 2) {
+            return null;
```
xstefank: returning null is generally a bad practice, especially if it is undocumented, the caller of this method can easily get NPE

src/main/java/org/xstefank/webhook/WebHookEndpoint.java
```diff
@@ -53,9 +57,12 @@ private static FormatConfig readConfig() {
         ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
 
         try {
-            return mapper.readValue(configFile, FormatConfig.class);
+            FormatConfig formatConfig = mapper.readValue(configFile, FormatConfig.class);
+            if (controlHandler.control(formatConfig))
```
xstefank: `if (isValidConfiguration(formatConfig))`

src/main/java/org/xstefank/verification/RepositoryFormatVerification.java
```diff
@@ -0,0 +1,21 @@
...
+
+import org.xstefank.model.yaml.FormatConfig;
+
+public class RepositoryFormatVerification implements Verification {
+
+
+    private final String REGEX = "^[A-z,0-9].*/[A-z,0-9,-].*$";
+    private final String MESSAGE = "Bad repository format in configuration file";
+
+    public boolean verify(FormatConfig formatConfig) {
```
xstefank: please add `@Override` - https://stackoverflow.com/questions/94361/when-do-you-use-javas-override-annotation-and-why

src/main/java/org/xstefank/verification/RepositoryFormatVerification.java
```diff
@@ -0,0 +1,21 @@
...
+        String repoFormat = formatConfig.getRepository();
+        if (repoFormat.matches(REGEX) && repoFormat.split("/").length == 2)
```
xstefank: this can still be abbreviated to one line

src/main/java/org/xstefank/verification/RepositoryFormatVerification.java
```diff
@@ -0,0 +1,21 @@
...
+            return true;
+        return false;
+    }
+
+    public String getErrorMessage() {
```
xstefank: `@Override`

src/main/java/org/xstefank/verification/Verification.java
```diff
@@ -0,0 +1,9 @@
+package org.xstefank.verification;
+
+import org.xstefank.model.yaml.FormatConfig;
+
+public interface Verification {
+
+    boolean verify(FormatConfig formatConfig);
+    String getErrorMessage();
```
xstefank: it would be sufficient to document here how `getErrorMessage` behaves in case `verify` wasn't called or the verification was successful, maybe it would be better to ask if it does make sense as a separate method

src/main/java/org/xstefank/verification/VerificationHandler.java
```diff
@@ -0,0 +1,32 @@
...
+                continue;
+            else
+                throw new BadFormatException(verification.getErrorMessage());
+        }
+    }
+
+    private List<Verification> registerVerifications() {
+        List<Verification> verifications = new ArrayList<>();
+
+        verifications.add(new RepositoryFormatVerification());
```
xstefank: can you come up with one or two more verifications so all this effort is justified please?

src/main/java/org/xstefank/webhook/WebHookEndpoint.java
```diff
@@ -7,21 +7,26 @@
...
 import java.io.File;
 import java.io.IOException;
+
 import static org.xstefank.check.TemplateChecker.TEMPLATE_FORMAT_FILE;
 
 @Path("/")
 public class WebHookEndpoint {
 
     private static final Logger log = Logger.getLogger(WebHookEndpoint.class);
+    private static VerificationHandler verificationHandler = new VerificationHandler();
```
xstefank: I don't think it's required to keep this handler in a field as its only used in one method, which is (I may be wrong here) called only once on the following line

src/main/java/org/xstefank/verification/VerificationHandler.java
```diff
@@ -0,0 +1,32 @@
...
+    private List<Verification> verifications;
+
+    public VerificationHandler() {
+        verifications = registerVerifications();
+    }
+
+    public void verifyConfiguration(FormatConfig formatConfig) throws BadFormatException {
+        for (Verification verification : verifications) {
+            if (verification.verify(formatConfig))
+                continue;
```
xstefank: why is this necessary here?

src/main/java/org/xstefank/verification/EmptyFormatVerification.java
```diff
@@ -0,0 +1,13 @@
...
+
+import org.xstefank.model.yaml.FormatConfig;
+
+public class EmptyFormatVerification implements Verification {
+
+    private final String MESSAGE = "Format in configuration file cannot be empty";
+
+    @Override
+    public String verify(FormatConfig formatConfig) {
+        return formatConfig.getFormat() != null ? "valid" : MESSAGE;
```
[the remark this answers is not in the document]
lukasburda: You mean that i should directly verify the config file and not FormatConfig which loads the Mapper ? Am I right?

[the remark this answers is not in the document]
xstefank: yes, however I think if you are able to declare the configuration without format element than this is a valid case, it would be worth while to check if the element can be made required somehow

src/main/java/org/xstefank/verification/RepositoryFormatVerification.java
```diff
@@ -0,0 +1,17 @@
...
+
+public class RepositoryFormatVerification implements Verification {
+
+    private final String MESSAGE = "Bad repository format in configuration file";
+    private final String REGEX = "^[A-z,0-9]*./[A-z,0-9,-].*$";
+
+    @Override
+    public void verify(FormatConfig formatConfig) throws BadFormatException {
+        String repoFormat = formatConfig.getRepository();
+        if (!repoFormat.matches(REGEX) || repoFormat.split("/").length != 2) {
```
xstefank: I am not sure `|| repoFormat.split("/").length != 2` is required

src/main/java/org/xstefank/verification/RepositoryFormatVerification.java
```diff
@@ -0,0 +1,17 @@
+package org.xstefank.verification;
+
+import org.xstefank.model.yaml.FormatConfig;
+
+public class RepositoryFormatVerification implements Verification {
+
+    private final String MESSAGE = "Bad repository format in configuration file";
+    private final String REGEX = "^[A-z,0-9]*./[A-z,0-9,-].*$";
```
xstefank: are you sure this regex is right? can you please include the link to the official rules for the repository naming conventions/rules?

src/main/java/org/xstefank/webhook/WebHookEndpoint.java
```diff
@@ -7,22 +7,26 @@
...
 import java.io.IOException;
+
 import static org.xstefank.check.TemplateChecker.TEMPLATE_FORMAT_FILE;
 
 @Path("/")
 public class WebHookEndpoint {
 
     private static final Logger log = Logger.getLogger(WebHookEndpoint.class);
-    private FormatConfig config = readConfig();
+    private static final FormatConfig config = readConfig();
```
xstefank: why this change?

src/main/java/org/xstefank/webhook/WebHookEndpoint.java
```diff
@@ -51,11 +55,12 @@ private static FormatConfig readConfig() {
         log.info(configFileName);
         File configFile = new File(configFileName);
         ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
-
         try {
-            return mapper.readValue(configFile, FormatConfig.class);
-        } catch (IOException e) {
+            FormatConfig formatConfig = mapper.readValue(configFile, FormatConfig.class);
+            new VerificationHandler(formatConfig).verifyConfiguration();
```
xstefank: looking at this line alone, wouldn't it be better to have a static method on the Verification handler that takes formatConfig as an argument?

src/main/java/org/xstefank/verification/RepositoryFormatVerification.java
xstefank: @lukasburda try to type `test/slash-repo` as a name of the new repository at GitHub ;)

xstefank: but I see now that you want to check whether what the user typed is a valid repository, still I am not convinced that this check is necessary and it cannot be caught directly by regex

src/main/java/org/xstefank/verification/RepositoryFormatVerification.java
```diff
@@ -0,0 +1,12 @@
+package org.xstefank.verification;
+
+import org.xstefank.model.yaml.FormatConfig;
+
+public class RepositoryFormatVerification implements Verification {
+
+    @Override
+    public void verify(FormatConfig formatConfig) throws IllegalConfigurationException {
+        if (!formatConfig.getRepository().matches("^[a-zA-Z0-9_]*./[a-zA-Z0-9_-]*.$"))
```
xstefank: what is the meaning of the dot character in front of the dollar sign please?

src/test/java/org/xstefank/verification/VerificationsTest.java
```diff
@@ -0,0 +1,32 @@
+package org.xstefank.verification;
+
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.xstefank.model.yaml.FormatConfig;
+
+import static org.xstefank.TestUtils.*;
```
xstefank: please don't use wildcard imports, there should be check for this in checkstyle

Issue #28, opened 2018-09-10
Add whitelisting capabilities


xstefank:
This enhancement would introduce a simple persistent storage of repository users who are eligible to do some advanced actions with their PRs. For instance, running CI without any admin confirmation. 
This issue will include:

- creating persistent user list
- adding option to include a user on this list dynamically
- adding a persistent admin list (users allowed to modify user list)
- add a simple way to include new actions when the user on user list creates / modifies the PR

PR #34, opened 2018-10-19
[Issue #28] Add whitelisting capabilities

lukasburda:
Issue: https://github.com/xstefank/tyr/issues/28


src/main/java/org/xstefank/model/WhitelistProcessing.java
```diff
@@ -0,0 +1,131 @@
...
+
+    private void processWithPrCommands(JsonNode payload, String authorUsername, String commUsername, String message, String commentId) {
+        if (message.matches(pullRequestCommands.getAddUserCommand()) && !LOGS.hasElement(commentId)
+                && ADMIN_LIST.hasElement(commUsername)) {
+            USER_LIST.addElement(authorUsername);
+            LOGS.addElement(commentId);
+            sendStatus(payload);
+        }
+
+        if (message.matches(pullRequestCommands.getRemoveUserCommand()) && !LOGS.hasElement(commentId)
```
xstefank: else if would be more sufficient

src/main/java/org/xstefank/model/WhitelistProcessing.java
```diff
@@ -0,0 +1,131 @@
...
+            USER_LIST.removeElement(authorUsername);
+            LOGS.addElement(commentId);
+        }
+
+        if (message.matches(pullRequestCommands.getRetestCommand()) && !LOGS.hasElement(commentId)
+                && ADMIN_LIST.hasElement(commUsername) && USER_LIST.hasElement(authorUsername)) {
+            LOGS.addElement(commentId);
+            sendStatus(payload);
+        }
+    }
```
xstefank: I don't like the approach you've taken here. The commands should be configurable, you've only hardcoded 3 possible commands so far. We would like to have an option to define for users some form of execution if the command is included

src/main/java/org/xstefank/model/WhitelistProcessing.java
```diff
@@ -0,0 +1,131 @@
...
+
+    private void processWithIssueCommands(String authorUsername, String commUsername, String message, String commentId) {
+        if (message.matches(issueCommands.getAddUserCommand()) && !LOGS.hasElement(commentId)
+                && ADMIN_LIST.hasElement(commUsername)) {
+            USER_LIST.addElement(authorUsername);
+            LOGS.addElement(commentId);
+        }
+
+        if (message.matches(issueCommands.getRemoveUserCommand()) && !LOGS.hasElement(commentId)
```
xstefank: do we really want to have whitelist capabilities on issues also?

src/main/java/org/xstefank/model/WhitelistProcessing.java
```diff
@@ -0,0 +1,131 @@
...
+                && ADMIN_LIST.hasElement(commUsername) && USER_LIST.hasElement(authorUsername)) {
+            USER_LIST.removeElement(authorUsername);
+            LOGS.addElement(commentId);
+        }
+    }
+
+    private void deleteLog(JsonNode payload) {
+        String id = payload.get(Utils.COMMENT).get(Utils.ID).asText();
+        if (LOGS.hasElement(id)) {
+            LOGS.removeElement(id);
```
xstefank: it won't remove the element if it is not present in the list ;)

src/main/java/org/xstefank/model/WhitelistProcessing.java
```diff
@@ -0,0 +1,131 @@
...
+        }
```
xstefank: I don't understand the point of this log, if you could explain please

src/main/java/org/xstefank/model/Utils.java
```diff
@@ -16,6 +17,10 @@
     public static final String TOKEN_ENV = "GITHUB_OAUTH_TOKEN";
     public static final String JBOSS_CONFIG_DIR = "jboss.server.config.dir";
     public static final String TEMPLATE_FORMAT_URL = "template.format.url";
+    public static final String USERLIST_FILE_NAME = "user-list";
+    public static final String ADMINLIST_FILE_NAME = "admin-list";
+    public static final String GITHUB_CACHE_FILE_NAME = "github.cache";
```
xstefank: I see this is later used as log, what is the point of this file?

src/main/java/org/xstefank/model/WhitelistProcessing.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    private void sendStatus(JsonNode payload) {
```
xstefank: this method does more than just sending status

src/main/java/org/xstefank/model/WhitelistProcessing.java
```diff
@@ -0,0 +1,131 @@
...
+            sendStatus(payload);
+        }
+
+        if (message.matches(pullRequestCommands.getRemoveUserCommand()) && !LOGS.hasElement(commentId)
+                && ADMIN_LIST.hasElement(commUsername) && USER_LIST.hasElement(authorUsername)) {
+            USER_LIST.removeElement(authorUsername);
+            LOGS.addElement(commentId);
+        }
+
+        if (message.matches(pullRequestCommands.getRetestCommand()) && !LOGS.hasElement(commentId)
```
xstefank: also we should distinguish later retesting tyr checks vs retesting CI

src/main/java/org/xstefank/model/yaml/IssueCommands.java
```diff
@@ -0,0 +1,23 @@
+package org.xstefank.model.yaml;
+
+public class IssueCommands {
```
xstefank: I would definitely prefer dynamic approach to this

src/main/java/org/xstefank/model/yaml/PullRequestCommands.java
```diff
@@ -0,0 +1,32 @@
+package org.xstefank.model.yaml;
+
+public class PullRequestCommands {
```
xstefank: same as above

src/main/java/org/xstefank/webhook/WebHookEndpoint.java
```diff
@@ -70,4 +65,37 @@ private static FormatConfig readConfig() {
...
     }
+
+    private void processWithWhitelisting(JsonNode payload) {
+        if (whitelistProcessing == null) {
+            whitelistProcessing = new WhitelistProcessing(templateChecker, config);
+        }
+        if (payload.has(Utils.PULL_REQUEST) && whitelistProcessing.isPrAuthorWhitelisted(payload)) {
+            sendStatus(payload);
+        } else if (payload.has(Utils.ISSUE)) {
+            whitelistProcessing.process(payload);
```
xstefank: in this call you are again unnecessarily checking if payload isn't a PR

src/main/java/org/xstefank/webhook/WebHookEndpoint.java
```diff
@@ -70,4 +65,37 @@ private static FormatConfig readConfig() {
...
+        if (!SkipCheck.shouldSkip(payload, config)) {
+            String errorMessage = templateChecker.checkPR(payload);
+            if (errorMessage != null) {
+                log.info("updating status");
+
+                GitHubAPI.updateCommitStatus(config.getRepository(),
+                        payload.get(Utils.HEAD).get(Utils.SHA).asText(),
+                        errorMessage.isEmpty() ? CommitStatus.SUCCESS : CommitStatus.ERROR,
+                        config.getStatusUrl(),
+                        errorMessage.isEmpty() ? "valid" : errorMessage, "PR format check");
```
xstefank: copy paste...

src/main/resources/admin-list
```diff
@@ -0,0 +1,2 @@
+# username
```
xstefank: rename file to admin-list.template or admin-list.example

src/main/resources/user-list
```diff
@@ -0,0 +1,2 @@
+# username
+# nextUsername
```
xstefank: same rename this file please

src/test/java/org/xstefank/TestUtils.java
```diff
@@ -29,10 +34,27 @@ public static FormatConfig loadFormatFromYamlFile(String fileName) {
...
+        return false;
+    }
+
+
     private static JsonNode loadJson(String fileName) {
         try {
             File file = new File(TestUtils.class.getClassLoader().getResource(fileName).getFile());
-            return new ObjectMapper().readTree(file);
+            JsonNode jNode = new ObjectMapper().readTree(file);
+            return Utils.reformatPayload(jNode);
```
xstefank: I am not sure about this. If something starts failing in `reformatPayload` a lot of not associated tests can start failing

src/test/resources/testConfig.properties
```diff
@@ -1,2 +1,3 @@
 github.oauth.token=351wa351d38aw4c97w98f7987ew98f987we97gs4
-template.format.url=https://www.someniceurlhere.org/
\ No newline at end of file
+template.format.url=https://www.someniceurlhere.org/
+whitelist.enabled=true
```
xstefank: tests that haven't been accounting for whitelisting shouldn't be affected (in other words you don't need whitelisting for all tests as it may be a performance bottleneck)

src/main/java/org/xstefank/api/TeamCityAPI.java
```diff
@@ -0,0 +1,81 @@
...
+        this.encryptedCredentials = encryptCredentials(Utils.getTyrProperty(Utils.TEAMCITY_USER_PROPERTY),
+                Utils.getTyrProperty(Utils.TEAMCITY_PASS_PROPERTY));
+
+        this.buildId = Utils.getTyrProperty(Utils.TEAMCITY_BUILD_CONFIG);
+    }
+
+    TeamCityAPI(String host, int port, String username, String password, String buildId) {
+        this.baseUrl = getBaseUrl(host, port);
+        this.encryptedCredentials = encryptCredentials(username, password);
+        this.buildId = buildId;
```
xstefank: Can you elaborate a little on this property? Does it mean that each instance of TeamCityAPI is associated with only one build?

src/main/java/org/xstefank/whitelist/WhitelistProcessing.java
```diff
@@ -0,0 +1,74 @@
...
+    private List<Command> getCommands(FormatConfig config) {
+        List<Command> commands = new ArrayList<>();
+        List<ContinuousIntegrationAPI> apis = loadApis();
+
+        if (config.getFormat().getCommands() != null ||
+                !config.getFormat().getCommands().isEmpty()) {
+
+            Map<String, String> regexMap = config.getFormat().getCommands();
+
+            String addUserCommandRegex = regexMap.get("AddUserCommand");
```
xstefank: prone to change and error, should be constant in the command itself?

src/test/java/org/xstefank/api/TeamCityAPITest.java
```diff
@@ -0,0 +1,12 @@
+package org.xstefank.api;
+
+import org.junit.Test;
+
+public class TeamCityAPITest {
+
+    @Test(expected = NullPointerException.class)
+    public void testTriggerBuildNullParams() {
+        TeamCityAPI teamCityAPI = new TeamCityAPI(null, 0, null, null, null);
```
xstefank: where is the NPE thrown? will the second line be ever executed?

src/test/java/org/xstefank/model/UserListTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+public class UserListTest {
+
+    private static final String USERNAME = "testUser";
+
+    private UserList testUserList;
+
+    @Before
+    public void before() {
+        testUserList = new UserList(TestUtils.RESOURCES_PATH, "test.list");
```
xstefank: what kind of file extension is list?

src/test/java/org/xstefank/model/UserListTest.java
```diff
@@ -0,0 +1,36 @@
...
+        testUserList.addUser(USERNAME);
+    }
+
+    @Test
+    public void testAddUserToList() {
+        Assert.assertTrue(TestUtils.isUserListed(testUserList, USERNAME));
```
xstefank: why is not this method in the UserList?

src/test/java/org/xstefank/utils/TestUtils.java
```diff
@@ -0,0 +1,90 @@
...
+                return trimJsonPullRequestElement(jNode);
+            } else {
+                return jNode;
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Cannot load file "+ fileName);
+        }
+    }
+
+    private static JsonNode trimJsonPullRequestElement(JsonNode payload) {
```
xstefank: trim?

src/test/java/org/xstefank/whitelist/RemoveUserCommandTest.java
```diff
@@ -0,0 +1,37 @@
...
+import org.junit.Test;
+import org.xstefank.utils.TestUtils;
+
+public class RemoveUserCommandTest {
+
+    private RemoveUserCommand removeUserCommand;
+
+    @Before
+    public void before() {
+        TestUtils.USER_LIST.addUser("prUser");
```
xstefank: not a good practice to modify public static final property in different class just for the test purposes

src/test/java/org/xstefank/whitelist/RetestCommandTest.java
```diff
@@ -0,0 +1,25 @@
...
+
+public class RetestCommandTest {
+
+    private RetestCommand retestCommand;
+
+    @Before
+    public void before() {
+        List<ContinuousIntegrationAPI> apis = new ArrayList<>();
+        String regex = TestUtils.FORMAT_CONFIG.getFormat().getCommands().get("RetestCommand");
+        retestCommand = new RetestCommand(regex, apis);
```
xstefank: why is this setup necessary if only test present tests NPE?

src/test/java/org/xstefank/utils/TestAPI.java
```diff
@@ -0,0 +1,12 @@
+package org.xstefank.utils;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import org.xstefank.api.ContinuousIntegrationAPI;
+
+public class TestAPI implements ContinuousIntegrationAPI {
```
xstefank: not being used anywhere?

src/main/java/org/xstefank/model/PersistentList.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+public class PersistentList {
+
+    private List<String> list = new ArrayList<>();
+    private File file;
+
+    public PersistentList(String dirName, String fileName) {
+        if (dirName == null && fileName == null) {
```
xstefank: this will allow to `dirName == null` and `fileName != null` which can cause NPE later

src/main/java/org/xstefank/model/PersistentList.java
```diff
@@ -0,0 +1,74 @@
...
+    public boolean hasUsername(String username) {
+        return list.contains(username);
+    }
+
+    public boolean isUserBackedInFile(String username) {
+        try (FileReader fileReader = new FileReader(file);
+             BufferedReader bufferedReader = new BufferedReader(fileReader)) {
+            String line = bufferedReader.readLine();
+            while (line != null) {
+                if (line.contains(username)) {
```
xstefank:
```suggestion
                if (line.equals(username)) {
```
??

src/main/java/org/xstefank/model/PersistentList.java
```diff
@@ -0,0 +1,74 @@
...
+        } catch (IOException e) {
+            throw new IllegalStateException("Cannot read file located in " + file.getPath(), e);
+        }
+        return false;
+    }
+
+    private void loadFileContent(File file) {
+        try (BufferedReader br = new BufferedReader(new FileReader(file))) {
+            String line = br.readLine();
+            while (line != null) {
```
xstefank: Would it be possible to do while (br.readLine() != null) { ... } ?

src/main/java/org/xstefank/model/Utils.java
```diff
@@ -24,12 +27,29 @@
...
     //Commit payload
     public static final String COMMIT = "commit";
     public static final String MESSAGE = "message";
     public static final String URL = "url";
 
+    // Teamcity API
+    public static final String TEAMCITY_HOST_PROPERTY = "teamcity.host";
+    public static final String TEAMCITY_PORT_PROPERTY = "teamcity.port";
+    public static final String TEAMCITY_USER_PROPERTY = "teamcity.user";
+    public static final String TEAMCITY_PASS_PROPERTY = "teamcity.password";
```
xstefank:
```suggestion
    public static final String TEAMCITY_PASSWORD_PROPERTY = "teamcity.password";
```

src/main/java/org/xstefank/model/Utils.java
```diff
@@ -59,4 +79,8 @@ public static URL getFormatUrl() throws MalformedURLException {
             target = getTyrProperty(TEMPLATE_FORMAT_URL);
         return target != null ? new URL(target) : null;
     }
+
+    public static boolean getBooleanProperty(String key) {
```
xstefank: maybe add also getIntProperty? you were parsing integer from getTyrProperty as port number

src/main/java/org/xstefank/webhook/WebHookEndpoint.java
```diff
@@ -28,23 +29,20 @@
...
-                GitHubAPI.updateCommitStatus(config.getRepository(),
-                        payload.get(Utils.PULL_REQUEST).get(Utils.HEAD).get(Utils.SHA).asText(),
-                        errorMessage.isEmpty() ? CommitStatus.SUCCESS : CommitStatus.ERROR,
-                        config.getStatusUrl(),
-                        errorMessage.isEmpty() ? "valid" : errorMessage, "PR format check");
-            }
+    public void processRequest(JsonNode payload) {
+        if (whitelistProcessing != null &&
+                payload.has(Utils.ISSUE)) {
+            whitelistProcessing.process(payload);
```
xstefank:
```suggestion
            whitelistProcessing.processComment(payload);
```
maybe this could better indicate that it's going to try to match comment commands?

src/main/java/org/xstefank/whitelist/AddUserCommand.java
```diff
@@ -0,0 +1,32 @@
...
+
+    private String commandRegex;
+
+    public AddUserCommand(String commandRegex) {
+        this.commandRegex = commandRegex;
+    }
+
+    @Override
+    public void process(JsonNode payload, PersistentList adminList, PersistentList userList) {
+        String authorUsername = WhitelistProcessing.getAuthorUsernameFromIssuePayload(payload);
```
xstefank: I'm not a big fan of very long names, I can see it's taking payload as argument ;)

src/main/java/org/xstefank/whitelist/WhitelistProcessing.java
```diff
@@ -0,0 +1,92 @@
...
+    public WhitelistProcessing(FormatConfig config) {
+        String dirName = System.getProperty(Utils.JBOSS_CONFIG_DIR);
+        userList = new PersistentList(dirName, Utils.USERLIST_FILE_NAME);
+        adminList = new PersistentList(dirName, Utils.ADMINLIST_FILE_NAME);
+        this.commands = getCommands(config);
+    }
+
+    public static boolean isUserEligibleForCIRun(JsonNode payload, PersistentList adminList) {
+        String username = payload.get(Utils.COMMENT).get(Utils.USER).get(Utils.LOGIN).asText();
+        return adminList.hasUsername(username);
```
xstefank: so only admins are eligible for running test CI?

src/main/java/org/xstefank/whitelist/WhitelistProcessing.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    public static String getAuthorUsernameFromIssuePayload(JsonNode payload) {
```
xstefank: maybe the last two methods can be combined? they are always being used together

src/main/java/org/xstefank/whitelist/WhitelistProcessing.java
```diff
@@ -0,0 +1,92 @@
...
+                    command.process(payload, adminList, userList);
+                }
+            }
+        }
+    }
+
+    private List<Command> getCommands(FormatConfig config) {
+        List<Command> commands = new ArrayList<>();
+
+        if (config.getFormat().getCommands() != null ||
```
xstefank: save `config.getFormat().getCommands()` prior to all these invocations, it is unnecessary to always call two methods to get to the same information

src/main/java/org/xstefank/whitelist/WhitelistProcessing.java
```diff
@@ -0,0 +1,92 @@
...
+                !config.getFormat().getCommands().isEmpty()) {
+
+            Map<String, String> regexMap = config.getFormat().getCommands();
+
+            String addUserCommandRegex = regexMap.get(AddUserCommand.NAME);
```
xstefank: maybe try to take here similar approach as it was taken with additional checks?

src/main/java/org/xstefank/whitelist/WhitelistProcessing.java
```diff
@@ -0,0 +1,92 @@
...
+
+            String retestCommandRegex = regexMap.get(RetestCommand.NAME);
+            if (retestCommandRegex != null) {
+                commands.add(new RetestCommand(retestCommandRegex));
+            }
+        }
+        return commands;
+    }
+
+    private static List<ContinuousIntegrationAPI> loadApis() {
```
xstefank: I still think that it is not really an API always ;)

src/test/java/org/xstefank/api/TeamCityAPITest.java
xstefank: maybe you can provide testing backing configuration file to avoid the need for this constructor? not necessary though

src/test/java/org/xstefank/model/PersistentListTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+    private final String USERNAME = "testUser";
+    private final String FILE_NAME = "testList.txt";
+
+    private PersistentList testUserList;
+
+    @Before
+    public void before() {
+        testUserList = new PersistentList(TestUtils.RESOURCES_PATH, FILE_NAME);
+        testUserList.addUser(USERNAME);
```
xstefank: cannot this line already fail?

src/test/java/org/xstefank/model/PersistentListTest.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    @Test
+    public void testAddUser() {
+        Assert.assertTrue(testUserList.isUserBackedInFile(USERNAME));
+    }
+
+    @Test
+    public void testHasUser() {
```
xstefank: it's really hard to see difference between two last tests

src/test/java/org/xstefank/whitelist/AddUserCommandTest.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void testProcessMethodNullParams() {
+        addUserCommand.process(null, null, null);
+    }
+
+    @After
+    public void after() {
+        TestUtils.deleteResourceFile(Utils.USERLIST_FILE_NAME);
```
xstefank: and adminlist?

src/main/java/org/xstefank/whitelist/WhitelistProcessing.java
lukasburda: There is difference in the statements in commands where these methods are used.

src/test/java/org/xstefank/whitelist/AddUserCommandTest.java
lukasburda: Tests reusing the adminlist everytime they are running.
  - Adminlist is always same. So it is not necessary creating it everytime the tests start.

src/main/java/org/xstefank/model/Utils.java
```diff
@@ -47,16 +67,24 @@ static void loadProperties(String dirName, String fileName) {
...
-    public static String getTyrProperty(String key) {
+    public static Properties getTyrProperties() {
         if (tyrProperties == null)
             loadProperties(System.getProperty(JBOSS_CONFIG_DIR), CONFIG_FILE);
-        return tyrProperties.getProperty(key);
+        return tyrProperties;
+    }
+
+    public static String getTyrProperty(String key) {
+        return getTyrProperties().getProperty(key);
```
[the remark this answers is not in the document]
lukasburda: Because of lazy instantiation of tyrProperties. If i accessed them directly I would implement here the lazy instantiation also.

src/main/java/org/xstefank/model/Utils.java
```diff
@@ -47,16 +67,24 @@ static void loadProperties(String dirName, String fileName) {
...
     }
 
     public static URL getFormatUrl() throws MalformedURLException {
         String target = System.getProperty(TEMPLATE_FORMAT_URL);
         if (target == null)
-            target = getTyrProperty(TEMPLATE_FORMAT_URL);
+            target = getTyrProperties().getProperty(TEMPLATE_FORMAT_URL);
```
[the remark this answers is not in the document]
lukasburda: like i meantion before.

src/main/java/org/xstefank/model/Utils.java
lukasburda: Ok I see what you mean. But it makes no sense to me write there that method because I passed all the properties to TeamCityCI class constructor. So the method would not be used anywhere for this case. Maybe for next force push it would...

src/main/java/org/xstefank/model/Utils.java
```diff
@@ -47,16 +67,24 @@ static void loadProperties(String dirName, String fileName) {
         }
     }
 
-    public static String getTyrProperty(String key) {
+    public static Properties getTyrProperties() {
         if (tyrProperties == null)
             loadProperties(System.getProperty(JBOSS_CONFIG_DIR), CONFIG_FILE);
-        return tyrProperties.getProperty(key);
+        return tyrProperties;
```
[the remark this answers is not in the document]
xstefank: gotcha, thanks

src/test/java/org/xstefank/ci/TeamCityCITest.java
```diff
@@ -0,0 +1,11 @@
+package org.xstefank.ci;
+
+import org.junit.Test;
+
+public class TeamCityCITest {
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testTriggerBuildNullParams() {
```
xstefank: there is no call to trigger CI?

src/test/java/org/xstefank/TestUtils.java
```diff
@@ -13,11 +15,13 @@
 
     public static final String YAML_DIR = "yaml";
     public static final String JSON_DIR = "json";
+    public static final String RESOURCES_PATH = "src/test/resources";
```
xstefank: this not a right place to create temporary test files

src/test/java/org/xstefank/whitelist/AddUserCommandTest.java
```diff
@@ -0,0 +1,58 @@
...
+        // Admin list is being reused so it's file won't be deleted
```
xstefank:
```suggestion
        // Admin list is being reused so its file won't be deleted
```

src/test/java/org/xstefank/TestUtils.java
```diff
@@ -29,6 +34,28 @@ public static FormatConfig loadFormatFromYamlFile(String fileName) {
         }
     }
 
+    public static void deleteResourceFile(String fileName) {
+        File file = new File(RESOURCES_PATH, fileName);
```
xstefank: still deleting files in `src/test/resources` ?

src/test/java/org/xstefank/TestUtils.java
```diff
@@ -29,6 +34,28 @@ public static FormatConfig loadFormatFromYamlFile(String fileName) {
...
+        if (file.exists()) {
+            file.delete();
+        }
+    }
+
+    public static boolean fileContainsElement(File file, String element) {
```
xstefank: rather fileContainsLine? It's not clear what is meant be element

src/test/java/org/xstefank/model/PersistentListTest.java
```diff
@@ -0,0 +1,37 @@
...
+public class PersistentListTest {
+
+    private final String USERNAME = "testUser";
+    private final String FILE_NAME = "testList.txt";
+
+    private PersistentList testUserList;
+
+    @Before
+    public void before() {
+        testUserList = new PersistentList(TestUtils.TARGET_DIR, FILE_NAME);
```
xstefank: creating in testList.txt in "basedir/target" but after is looking for it to delete it in "src/test/resources" ?

src/test/java/org/xstefank/whitelist/AddUserCommandTest.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void testProcessMethodNullParams() {
+        addUserCommand.process(null, null, null, null);
+    }
+
+    @After
+    public void after() {
+        TestUtils.deleteResourceFile(Utils.USERLIST_FILE_NAME);
```
xstefank: Again will not find it to close it as it is created in the target dir?

src/test/java/org/xstefank/whitelist/RetestCommandTest.java
```diff
@@ -0,0 +1,28 @@
...
+
+import static org.powermock.api.support.membermodification.MemberMatcher.method;
+
+@RunWith(PowerMockRunner.class)
+@PrepareForTest(WhitelistProcessing.class)
+public class RetestCommandTest {
+
+    @Before
+    public void before() {
+        PowerMockito.stub(method(WhitelistProcessing.class, "loadCIs"))
```
xstefank: is this necessary now that CIs are configurable?

lukasburda: I've just tried what you asked for.. But I think it does not make sense at now.
https://github.com/xstefank/tyr/pull/34#discussion_r238159236

src/main/java/org/xstefank/whitelist/RetestCommand.java
```diff
@@ -0,0 +1,35 @@
...
+    public static final String NAME = "RetestCommand";
+
+    private String commandRegex;
+
+    @Override
+    public void process(JsonNode payload, PersistentList adminList,
+                        PersistentList userList, List<ContinuousIntegration> CIs) {
+        String authorUsername = WhitelistProcessing.getAuthorUsername(payload);
+
+        if (WhitelistProcessing.isCommentAuthorAdministrator(payload, adminList) &&
```
xstefank: retest should be available to all enlisted users (in the user list), no need for the comment author to be admin

src/main/java/org/xstefank/whitelist/WhitelistProcessing.java
xstefank: still valid I think

src/main/java/org/xstefank/whitelist/WhitelistProcessing.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    static void triggerCI(JsonNode payload, List<ContinuousIntegration> CIs) {
+        JsonNode prJson = GitHubAPI.getJsonWithPullRequest(payload);
+        for (ContinuousIntegration ci : CIs) {
+            ci.triggerBuild(prJson);
+        }
+    }
+
+    public void processComment(JsonNode payload) {
```
xstefank: maybe just to be totally clear - processPRComment?

src/test/java/org/xstefank/TestUtils.java
```diff
@@ -29,6 +34,27 @@ public static FormatConfig loadFormatFromYamlFile(String fileName) {
         }
     }
 
+    public static void deleteTempFile(File file) {
+        if (file.exists()) {
+            file.delete();
+        }
+    }
+
+    public static boolean fileContainsLine(File file, String element) {
```
xstefank:
```suggestion
    public static boolean fileContainsLine(File file, String line) {
```

src/test/java/org/xstefank/model/PersistentListTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+    private final String USERNAME = "testUser";
+    private final String FILE_NAME = "testList.txt";
+
+    private File userListFile = new File(TestUtils.TARGET_DIR, FILE_NAME);
+    private PersistentList testUserList;
+
+    @Before
+    public void before() {
+        testUserList = new PersistentList(TestUtils.TARGET_DIR, FILE_NAME);
```
xstefank: I don't think this is doing what you've intended - you create userListFile on line 15, and create PersistentList from it before every test. But `@After` deletes userListFile after first test so second test will create a new file at the same place by the call to PersistentList constructor - it's just no consistent so please clarify if this is what you've intended

src/test/java/org/xstefank/whitelist/AddUserCommandTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+@RunWith(PowerMockRunner.class)
+@PrepareForTest(GitHubAPI.class)
+public class AddUserCommandTest {
+
+    private AddUserCommand addUserCommand;
+    private PersistentList userList = new PersistentList(TestUtils.TARGET_DIR, Utils.USERLIST_FILE_NAME);
+    private PersistentList adminList = new PersistentList(TestUtils.RESOURCES_PATH, Utils.ADMINLIST_FILE_NAME);
+
+    private File userListFile = new File(TestUtils.TARGET_DIR, Utils.USERLIST_FILE_NAME);
```
xstefank: not necessary to keep it as a field if you are using it just for the removal

src/main/java/org/xstefank/model/yaml/Format.java
```diff
@@ -10,6 +12,10 @@
     private SkipPatterns skipPatterns;
     private Description description;
     private List<String> additional;
+    private Map<String, String> commands;
+
+    @JsonProperty("CI")
```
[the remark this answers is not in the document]
lukasburda: This is required if you want to write "CI" instead of "ci" to format.yaml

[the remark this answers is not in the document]
xstefank: thanks

src/test/java/org/xstefank/whitelist/RetestCommandTest.java
```diff
@@ -0,0 +1,12 @@
+package org.xstefank.whitelist;
+
+import org.junit.Test;
+
+public class RetestCommandTest {
```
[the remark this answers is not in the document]
lukasburda: Do you think that it is good idea to test this also for AddUserCommand?

src/main/java/org/xstefank/model/PersistentList.java
```diff
@@ -0,0 +1,67 @@
...
+                file.createNewFile();
+            } catch (IOException e) {
+                throw new IllegalStateException("Cannot create new file with name " + file.getName(), e);
+            }
+        } else {
+            loadFileContent(file);
+        }
+    }
+
+    public PersistentList(File file) {
```
xstefank: you can reuse one of these constructors for the other one

src/main/java/org/xstefank/whitelist/RetestCommand.java
```diff
@@ -0,0 +1,35 @@
...
+    @Override
+    public void process(JsonNode payload, PersistentList adminList,
+                        PersistentList userList) {
+        String pullRequestAuthor = WhitelistProcessing.getPRAuthor(payload);
+        String commentAuthor = WhitelistProcessing.getCommentAuthor(payload);
+
+        if (userList.hasUsername(pullRequestAuthor) &&
+                userList.hasUsername(commentAuthor) ||
+                adminList.hasUsername(commentAuthor) &&
+                userList.hasUsername(pullRequestAuthor)) {
```
xstefank: `userList.hasUsername(pullRequestAuthor)` needs to be true in both cases, I think you can rewrite this if 
 clause in a simpler way

src/test/java/org/xstefank/whitelist/AddUserCommandTest.java
```diff
@@ -0,0 +1,79 @@
...
+    @Test
+    public void testAddUserToList() {
+       addUserCommand.process(TestUtils.ISSUE_PAYLOAD, adminList, userList);
+        Assert.assertTrue(TestUtils.fileContainsLine(userListFile,"prUser"));
+    }
+
+    @Test
+    public void testTriggerBuildMethodCall() {
+        String prAuthor = TestUtils.ISSUE_PAYLOAD.get(Utils.ISSUE).get(Utils.USER).get(Utils.LOGIN).asText();
+        userList.addUser(prAuthor);
```
xstefank: I may be missing something, but isn't add user command doing precisely this?

src/test/java/org/xstefank/whitelist/AddUserCommandTest.java
```diff
@@ -0,0 +1,79 @@
...
+        addUserCommand.process(TestUtils.ISSUE_PAYLOAD, adminList, userList);
+        Assert.assertTrue(testCI.isTriggered());
+    }
+
+    @Test
+    public void testCommandRegexValue() {
+        String regex = TestUtils.FORMAT_CONFIG.getFormat().getCommands().get(AddUserCommand.NAME);
+        addUserCommand.setCommandRegex(regex);
+        Assert.assertEquals(regex, addUserCommand.getCommandRegex());
```
xstefank: this test tests only that if you set value with setter you will get the same back with getter? There is nothing wrong with the test, it's just a little unnecessary

src/test/java/org/xstefank/whitelist/RetestCommandTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+        retestCommand.process(TestUtils.ISSUE_PAYLOAD, adminList, userList);
+        Assert.assertTrue(testCI.isTriggered());
+    }
+
+    @Test
+    public void testCommandRegexValue() {
+        String regex = TestUtils.FORMAT_CONFIG.getFormat().getCommands().get(RetestCommand.NAME);
+        retestCommand.setCommandRegex(regex);
+        Assert.assertEquals(regex, retestCommand.getCommandRegex());
```
xstefank: see setter / getter comment above

src/main/java/org/xstefank/model/PersistentList.java
```diff
@@ -0,0 +1,67 @@
...
+        this.file = file;
+        if (file.exists()) {
+            loadFileContent(file);
+        }
```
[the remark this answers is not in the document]
lukasburda: I can call "createNewFile" method but the FileWriter I am using in "addUser" method, automatically creates new file if it hasn't existed when is called. It is working as you would expect. The method "hasUsername" can be called without created file because that file is not used there.

lukasburda:
retest this please

lukasburda:
ok to test

lukasburda:
retest this please

src/main/java/org/xstefank/model/PersistentList.java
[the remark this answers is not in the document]
xstefank: so what is the point of other constructor creating file if it does not exist?

src/main/java/org/xstefank/whitelist/WhitelistProcessing.java
```diff
@@ -0,0 +1,111 @@
...
+            String message = payload.get(Utils.COMMENT).get(Utils.BODY).asText();
+            for (Command command : commands) {
+                if (message.matches(command.getCommandRegex())) {
+                    command.process(payload, adminList, userList);
+                }
+            }
+        }
+    }
+
+    public void triggerCIOnOpened(JsonNode payload) {
```
xstefank: it should be clear here that payload is different than the one that is being passed to triggerCI - I don't really see the need for two methods

src/main/resources/user-list.txt.example
```diff
@@ -0,0 +1,2 @@
+# username
```
xstefank: is there really need to have two user-list admin-list example files with the same content?

src/test/java/org/xstefank/model/PersistentListTest.java
```diff
@@ -0,0 +1,45 @@
...
+    private PersistentList testUserList;
+
+    @BeforeClass
+    public static void beforeClass() {
+        TestUtils.deleteFileIfExists(userListFile);
+    }
+
+    @Before
+    public void before() {
+        testUserList = new PersistentList(userListFile);
```
xstefank: maybe you can add a test that asserts that file is created?

xstefank: and that contents of non-empty file are loaded successfully?

src/test/java/org/xstefank/whitelist/AddUserCommandTest.java
```diff
@@ -0,0 +1,70 @@
...
+    private static final TestCI testCI = new TestCI();
+
+    private final PersistentList userList = new PersistentList(userListFile);
+    private final PersistentList adminList = new PersistentList(TestUtils.RESOURCES_PATH, Utils.ADMINLIST_FILE_NAME);
+    private AddUserCommand addUserCommand;
+
+    @BeforeClass
+    public static void beforeClass() {
+        TestUtils.deleteFileIfExists(userListFile);
+        PowerMockito.stub(method(CILoader.class, "getCI", String.class))
```
xstefank: I don't understand why `getCI` method needs to be stubbed, maybe it's a little too much :) would it be possible to stub method that calls this? if not than it's ok

src/test/java/org/xstefank/whitelist/AddUserCommandTest.java
```diff
@@ -0,0 +1,70 @@
...
+                .toReturn(testCI);
+        WhitelistProcessing.loadCIs(TestUtils.FORMAT_CONFIG);
+    }
+
+    @Before
+    public void before() {
+        PowerMockito.stub(method(GitHubAPI.class, "getJsonWithPullRequest", JsonNode.class))
```
xstefank: why this stub is in before and the other is in beforeclass? if you can avoid loading this before each test it would be preferable

src/test/java/org/xstefank/whitelist/AddUserCommandTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Test
+    public void testAddUserToList() {
+       addUserCommand.process(TestUtils.ISSUE_PAYLOAD, adminList, userList);
+        Assert.assertTrue(TestUtils.fileContainsLine(userListFile,"prUser"));
+    }
+
+    @Test
+    public void testTriggerBuildMethodCall() {
+        addUserCommand.process(TestUtils.ISSUE_PAYLOAD, adminList, userList);
```
xstefank: this call is repeated in previous test, merge them into one please

src/test/resources/yaml/testTemplate.yaml
```diff
@@ -11,3 +11,6 @@ format:
   additional:
     - TitleIssueLinkIncludedCheck
     - OneCommitOnlyCheck
+
+  CI:
+    - TestCI
```
xstefank: is this necessary when you stub `getCI` method?

src/main/java/org/xstefank/webhook/WebHookEndpoint.java
```diff
@@ -28,23 +29,21 @@
...
-                        payload.get(Utils.PULL_REQUEST).get(Utils.HEAD).get(Utils.SHA).asText(),
-                        errorMessage.isEmpty() ? CommitStatus.SUCCESS : CommitStatus.ERROR,
-                        config.getStatusUrl(),
-                        errorMessage.isEmpty() ? "valid" : errorMessage, "PR format check");
-            }
+    public void processRequest(JsonNode payload) {
+        if (whitelistProcessing != null) {
+            processPRWithWhitelisting(payload);
+        } else if (payload.has(Utils.PULL_REQUEST)) {
+            processPullRequest(payload);
```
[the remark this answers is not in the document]
lukasburda: It works as it did

src/test/java/org/xstefank/whitelist/AddUserCommandTest.java
lukasburda: I think it makes sense because instead of this i would need to stub the loadCIs method that adds all CI to list which is used for triggering them.

src/test/java/org/xstefank/whitelist/AddUserCommandTest.java
lukasburda: I wanted to put the stub of this to beforeClass but it didn't work that way. I am using this stub for two tests in this class. So I am not sure if this is the right place for it.

src/test/resources/yaml/testTemplate.yaml
lukasburda: CI element must be specified.

xstefank: the question that arise whether it would be possible to substitu which CIs are available for configuration in the first place. That is to change from which options user can chose CIs from in the test

src/test/java/org/xstefank/ci/TestCI.java
```diff
@@ -0,0 +1,31 @@
...
+    private static final String NAME = "TestCI";
+    private boolean triggered;
+
+    public static TestCI getTestCIFromLoader() {
+        return (TestCI) CILoader.getCI(NAME);
+    }
+
+    public static void putTestCIToLoader() {
+        CILoader.addCI(NAME, new TestCI());
+    }
```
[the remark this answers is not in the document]
lukasburda: Well thanks for the question.. I think I need just this method because the method "addCI" is package private and I can't call it from "CommandsTest" class. Maybe I can try to refactor Commands to different package if it does make sense?

src/test/java/org/xstefank/whitelist/CommandsTest.java
```diff
@@ -0,0 +1,92 @@
...
+public class CommandsTest {
+
+    private static final String COMMENT_USER = "commentUser";
+    private static final String PR_AUTHOR = "prUser";
+
+    private static File userListFile;
+    private static File adminListFile;
+
+    private final AddUserCommand addUserCommand = new AddUserCommand();
+    private final RetestCommand retestCommand =  new RetestCommand();
```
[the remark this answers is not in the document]
lukasburda: If I separate these two commands. Would it make sense?

PR #36, opened 2018-12-04
[Issue #35] Properties should not be stored in Utils class

lukasburda:
Issue: https://github.com/xstefank/tyr/issues/35

### DIFF
--- a/src/main/java/org/xstefank/model/TyrProperties.java
+++ b/src/main/java/org/xstefank/model/TyrProperties.java
@@ -1,0 +1,36 @@
+package org.xstefank.model;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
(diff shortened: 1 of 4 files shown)


src/main/java/org/xstefank/model/TyrProperties.java
```diff
@@ -0,0 +1,36 @@
...
+
+import static org.xstefank.model.Utils.JBOSS_CONFIG_DIR;
+import static org.xstefank.model.Utils.CONFIG_FILE;
+
+public class TyrProperties {
+
+    private static final Properties properties =
+            loadProperties(System.getProperty(JBOSS_CONFIG_DIR), CONFIG_FILE);
+
+    public static String getStringProperty(String key) {
```
xstefank: also here `getProperty` should be sufficient enough right? user can see the return type

Issue #32, opened 2018-09-27
Code cleanup


xstefank:
mostly logging and redundant blocks

PR #41, opened 2019-01-24
[Issue #32] Code cleanup

lukasburda:
Issue: https://github.com/xstefank/tyr/issues/32

### DIFF
--- a/src/main/java/org/xstefank/check/TemplateChecker.java
+++ b/src/main/java/org/xstefank/check/TemplateChecker.java
@@ -18,6 +18,7 @@ package org.xstefank.check;
 import com.fasterxml.jackson.databind.JsonNode;
 import org.jboss.logging.Logger;
 import org.xstefank.check.additional.AdditionalChecks;
+import org.xstefank.model.Utils;
 import org.xstefank.model.yaml.FormatConfig;
 import org.xstefank.model.yaml.Format;
 import java.util.ArrayList;
@@ -38,8 +39,8 @@ public class TemplateChecker {
     }
 
     public String checkPR(JsonNode payload) {
-        log.info("checking PR");
-        String description = "";
+        log.debug("checking PR" + Utils.LINE_SEPARATOR + payload);
+        String errorMessage = "";
 
         if (checks.isEmpty()) {
             log.warn("No checks were requested in the configuration");
@@ -47,12 +48,12 @@ public class TemplateChecker {
         for (Check check : checks) {
             String message = check.check(payload);
             if (message != null) {
-                description = message;
+                errorMessage = message;
                 break;
             }
         }
 
-        return description;
+        return errorMessage;
     }
 
     private static List<Check> registerChecks(Format format) {
@@ -70,8 +71,8 @@ public class TemplateChecker {
             checks.add(new LatestCommitCheck(format.getCommit()));
         }
 
-        if (format.getAdditional() != null) {
-            for (String additional : format.getAdditional()) {
+        if (format.getAdditionalChecks() != null) {
+            for (String additional : format.getAdditionalChecks()) {
                 checks.add(AdditionalChecks.findCheck(additional));
             }
         }
--- a/src/main/java/org/xstefank/model/PersistentList.java
+++ b/src/main/java/org/xstefank/model/PersistentList.java
@@ -20,24 +20,30 @@ import java.io.File;
 import java.io.FileReader;
 import java.io.FileWriter;
 import java.io.IOException;
+import java.io.PrintWriter;
 import java.util.ArrayList;
-import java.util.List;
+import java.util.Collection;
+import java.util.Comparator;
+import java.util.function.UnaryOperator;
+import org.jboss.logging.Logger;
 
-public class PersistentList {
+public class PersistentList extends ArrayList<String> {
 
-    private List<String> list = new ArrayList<>();
+    private static final Logger log = Logger.getLogger(PersistentList.class);
+
     private File file;
 
     public PersistentList(File file) {
+        super();
         this.file = file;
         if (file.exists()) {
             loadFileContent(file);
-        } else {
-            try {
-                file.createNewFile();
-            } catch (IOException e) {
-                throw new IllegalStateException("Cannot create new file with name " + file.getName(), e);
-            }
+            return;
+        }
+        try {
+            file.createNewFile();
+        } catch (IOException e) {
+            throw new IllegalStateException("Cannot create new file with name " + file.getName(), e);
         }
     }
 
@@ -45,20 +51,74 @@ public class PersistentList {
         this(new File(dirName, fileName));
     }
 
-    public void addUser(String username) {
-        if (list.contains(username)) {
-            return;
+    @Override
+    public boolean add(String s) {
+        if (!super.add(s)) {
+            return false;
         }
-        list.add(username);
-        try (FileWriter fw = new FileWriter(file, true)) {
-            fw.write(username + "\n");
+        try (PrintWriter pw = new PrintWriter(new FileWriter(file, true))) {
+            pw.println(s);
         } catch (IOException e) {
-            throw new IllegalStateException("Cannot write user to file", e);
+            log.error(e);
+            super.remove(s);
+            return false;
         }
+        return true;
     }
 
-    public boolean hasUsername(String  username) {
-        return list.contains(username);
+    @Override
+    public void add(int index, String element) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public boolean addAll(Collection<? extends String> c) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public boolean addAll(int index, Collection<? extends String> c) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public void clear() {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public String remove(int index) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public boolean remove(Object o) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public boolean removeAll(Collection<?> c) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public void replaceAll(UnaryOperator<String> operator) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public boolean retainAll(Collection<?> c) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public String set(int index, String element) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public void sort(Comparator<? super String> c) {
+        throw new UnsupportedOperationException();
     }
 
     private void loadFileContent(File file) {
@@ -66,11 +126,11 @@ public class PersistentList {
             String line;
             while ((line = br.readLine()) != null) {
                 if (!line.startsWith("#")) {
-                    list.add(line);
+                    super.add(line);
                 }
             }
         } catch (IOException e) {
-            throw new IllegalStateException("Cannot read file " + file.getPath(), e);
+            throw new IllegalStateException("Cannot access file " + file.getPath(), e);
         }
     }
 }
--- a/src/main/java/org/xstefank/verification/InvalidConfigurationException.java
+++ b/src/main/java/org/xstefank/verification/InvalidConfigurationException.java
@@ -17,7 +17,7 @@ package org.xstefank.verification;
 
 public class InvalidConfigurationException extends Exception {
 
-    public InvalidConfigurationException(String message) {
+    InvalidConfigurationException(String message) {
         super(message);
     }
 }
--- a/src/main/java/org/xstefank/whitelist/Command.java
+++ b/src/main/java/org/xstefank/whitelist/Command.java
@@ -17,11 +17,30 @@ package org.xstefank.whitelist;
 
 import com.fasterxml.jackson.databind.JsonNode;
 
-public interface Command {
+/**
+ * Command that can process issue payload received from GitHub API
+ */
+public abstract class Command {
 
-    void process(JsonNode payload, WhitelistProcessing whitelistProcessing);
+    private String commandRegex;
 
-    String getCommandRegex();
+    /**
+     * This method is used for command processing.
+     * You can define your own functionality of new command
+     * that you want to implement. Method is triggered automatically
+     * when the command regex is matched. You can set the regex
+     * in {@link CommandsLoader} class and use it in format.yaml.
+     * @param payload json received from GitHub API.
+     * @param whitelistProcessing class that offers some useful functionality
+     *                            for working with Continuous Integration.
+     */
+    public abstract void process(JsonNode payload, WhitelistProcessing whitelistProcessing);
 
-    void setCommandRegex(String commandRegex);
+    String getCommandRegex() {
+        return commandRegex;
+    }
+
+    void setCommandRegex(String commandRegex) {
+        this.commandRegex = commandRegex;
+    }
 }
--- a/src/main/java/org/xstefank/whitelist/WhitelistProcessing.java
+++ b/src/main/java/org/xstefank/whitelist/WhitelistProcessing.java
@@ -19,6 +19,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import org.jboss.logging.Logger;
 import org.xstefank.ci.CILoader;
 import org.xstefank.ci.ContinuousIntegration;
 import org.xstefank.model.PersistentList;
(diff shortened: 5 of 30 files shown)


src/main/java/org/xstefank/whitelist/WhitelistProcessing.java
```diff
@@ -87,25 +90,31 @@ void addUserToUserList(String username) {
             if (command != null) {
                 command.setCommandRegex(regexMap.get(key));
                 commands.add(command);
+            } else {
+                log.warn("Command" + IDENTIFIED_WITH + key + DOES_NOT_EXISTS);
```
xstefank: you don't need two private constants for this. Try `log.warnf()`

src/main/java/org/xstefank/whitelist/WhitelistProcessing.java
```diff
@@ -87,25 +90,31 @@ void addUserToUserList(String username) {
...
             }
         }
 
         return commands;
     }
 
-    private void loadCIs(FormatConfig config) {
+    private List<ContinuousIntegration> loadCIs(FormatConfig config) {
+        List<ContinuousIntegration> tempCIList = new ArrayList<>();
```
xstefank: resultList

src/test/java/org/xstefank/check/RequiredRowsCheckTest.java
```diff
@@ -31,13 +30,15 @@ public void before() {
 
     @Test
     public void checkSimpleRegexMatch() {
-        Assert.assertNull("Cannot match valid description", requiredRowsCheck.check(TestUtils.TEST_PAYLOAD));
+        Assert.assertNull("Cannot match valid description",
+                requiredRowsCheck.check(TestUtils.TEST_PAYLOAD));
```
xstefank: I don't find the original line that long, but ok

src/test/java/org/xstefank/model/PersistentListTest.java
```diff
@@ -16,7 +15,7 @@
     private static File userListFile = new File(TestUtils.TARGET_DIR, FILE_NAME);
 
     private final String USERNAME = "testUser";
-    private PersistentList testUserList;
+    private List<String> testUserList;
```
xstefank: here you can keep persistenList as this unit test is testing precisely PersistentList

src/main/java/org/xstefank/model/PersistentList.java
```diff
@@ -30,32 +33,126 @@ public PersistentList(String dirName, String fileName) {
...
+    public boolean addAll(int index, Collection<? extends String> c) {
+        boolean changed = super.addAll(index, c);
+        writeListContentToFile();
+        return changed;
+    }
+
+    @Override
+    public boolean addAll(Collection<? extends String> c) {
+        boolean changed = super.addAll(c);
+        try (FileWriter fw = new FileWriter(file, true)){
```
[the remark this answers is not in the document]
lukasburda: Because if you use this method it just adds content of collection "c" after the last element in list. So It is safe to write it to file this way. I don't need to rewrite whole file.

src/main/java/org/xstefank/model/PersistentList.java
```diff
@@ -30,32 +34,136 @@ public PersistentList(String dirName, String fileName) {
...
+    @Override
+    public boolean add(String s) {
+        if (contains(s)) {
+            return false;
         }
-        list.add(username);
         try (FileWriter fw = new FileWriter(file, true)) {
-            fw.write(username + "\n");
+            fw.write(s + "\n");
+            return super.add(s);
```
xstefank: you should stay consistent with ordering of operations. Everywhere else you first modify in-memory list and then write file except here

src/main/java/org/xstefank/model/PersistentList.java
```diff
@@ -30,32 +34,136 @@ public PersistentList(String dirName, String fileName) {
...
+        return removed;
+    }
+
+    @Override
+    public void clear() {
+        super.clear();
+        try (FileWriter fw = new FileWriter(file, false)) {
+            fw.write("");
+        } catch (IOException e) {
+            throw new IllegalStateException("Cannot delete list elements", e);
```
xstefank: Cannot clear list

src/main/java/org/xstefank/model/PersistentList.java
```diff
@@ -30,32 +34,136 @@ public PersistentList(String dirName, String fileName) {
...
+
+    @Override
+    protected void removeRange(int fromIndex, int toIndex) {
+        super.removeRange(fromIndex, toIndex);
+        rewriteFileContent(fromIndex);
+    }
+
+    @Override
+    public boolean retainAll(Collection<?> c) {
+        throw new UnsupportedOperationException();
```
xstefank: why UnsupportedOperationException? and everywhere else

src/main/java/org/xstefank/model/PersistentList.java
```diff
@@ -30,32 +34,136 @@ public PersistentList(String dirName, String fileName) {
...
             throw new IllegalStateException("Cannot read file " + file.getPath(), e);
         }
     }
-}
+
+    private void rewriteFileContent(int fromIndex) {
+        try (RandomAccessFile randomAccessFile =
+                     new RandomAccessFile(file.getAbsolutePath(), "rw")) {
+            if (randomAccessFile.length() > 0) {
+                int fileEnd = 0;
```
xstefank: this should not be names as fileEnd. It is a temporal file end when you find fromIndex but the name is misleading.

src/main/java/org/xstefank/model/PersistentList.java
```diff
@@ -30,32 +34,136 @@ public PersistentList(String dirName, String fileName) {
...
+                for (int i = 0; i < fromIndex; i++) {
+                    fileEnd += get(i).length() + 1;
+                }
+                randomAccessFile.setLength(fileEnd);
+                randomAccessFile.seek(fileEnd);
+            }
+            for (int i = fromIndex; i < size(); i++) {
+                randomAccessFile.writeBytes(get(i) + "\n");
```
xstefank: this should be OS independent line separator

src/main/java/org/xstefank/model/PersistentList.java
```diff
@@ -30,32 +34,136 @@ public PersistentList(String dirName, String fileName) {
...
+            }
+        } catch (IOException e) {
+            e.printStackTrace();
```
xstefank: log exception, don't print stacktrace

src/main/java/org/xstefank/verification/InvalidConfigurationException.java
```diff
@@ -2,7 +2,9 @@
 
 public class InvalidConfigurationException extends Exception {
 
-    public InvalidConfigurationException(String message) {
+    private static final long serialVersionUID = 1L;
```
xstefank: https://stackoverflow.com/questions/888335/why-generate-long-serialversionuid-instead-of-a-simple-1l
http://frequal.com/java/PracticalSerialVersionIdGuidelines.html

src/main/java/org/xstefank/whitelist/WhitelistProcessing.java
```diff
@@ -63,49 +64,55 @@ String getPRAuthor(JsonNode issuePayload) {
...
+        return userList.contains(username);
     }
 
     void addUserToUserList(String username) {
-        userList.addUser(username);
+        userList.add(username);
     }
 
     private List<Command> getCommands(FormatConfig config) {
-        List<Command> commands = new ArrayList<>();
```
xstefank: why not commands?

src/main/java/org/xstefank/whitelist/WhitelistProcessing.java
```diff
@@ -63,49 +64,55 @@ String getPRAuthor(JsonNode issuePayload) {
...
             }
         }
 
-        return commands;
+        return resultList;
     }
 
-    private void loadCIs(FormatConfig config) {
+    private List<ContinuousIntegration> loadCIs(FormatConfig config) {
+        List<ContinuousIntegration> resultList = new ArrayList<>();
```
xstefank: why not continousIntegrations?

src/main/java/org/xstefank/check/TemplateChecker.java
```diff
@@ -23,21 +23,21 @@ public TemplateChecker(FormatConfig config) {
     }
 
     public String checkPR(JsonNode payload) {
-        log.info("checking PR");
-        String description = "";
+        log.debug("checking PR\n" + payload);
```
xstefank: @lukasburda you've filed windows issue lately. I believe this end of line is not going to work there

src/main/java/org/xstefank/model/PersistentList.java
```diff
@@ -5,15 +5,17 @@
...
+import java.io.RandomAccessFile;
 import java.util.ArrayList;
-import java.util.List;
+import java.util.Collection;
 
-public class PersistentList {
+public class PersistentList extends ArrayList<String> {
 
-    private List<String> list = new ArrayList<>();
+    private static final String LINE_SEPARATOR = System.getProperty("line.separator");
```
xstefank: https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#lineSeparator--

src/main/java/org/xstefank/model/PersistentList.java
```diff
@@ -30,32 +32,122 @@ public PersistentList(String dirName, String fileName) {
...
+    public boolean addAll(Collection<? extends String> c) {
+        boolean changed = super.addAll(c);
         try (FileWriter fw = new FileWriter(file, true)) {
-            fw.write(username + "\n");
+            for (String element : c) {
+                fw.write(element);
+            }
         } catch (IOException e) {
-            throw new IllegalStateException("Cannot write user to file", e);
+            throw new IllegalStateException("Cannot write elements to list", e);
```
xstefank: shouldn't we guard here against possible inconsistencies? (also maybe in some other methods) -- If collection c contains 3 elements, we successfully write first 2 on line 99 and then IOException is thrown then super has all 3 elements while the file has only 2

src/main/java/org/xstefank/model/PersistentList.java
```diff
@@ -30,32 +32,122 @@ public PersistentList(String dirName, String fileName) {
...
+                    characterCount += get(i).length() + LINE_SEPARATOR.length();
+                }
+            }
+            randomAccessFile.setLength(characterCount);
+            randomAccessFile.seek(characterCount);
+            for (int i = fromIndex; i < size(); i++) {
+                randomAccessFile.writeBytes(get(i) + LINE_SEPARATOR);
+            }
+        } catch (IOException e) {
+            throw new IllegalStateException("Cannot rewrite file content", e);
```
xstefank: same potential inconsistency problem

src/main/java/org/xstefank/whitelist/Command.java
```diff
@@ -2,11 +2,17 @@
...
 import com.fasterxml.jackson.databind.JsonNode;
 
-public interface Command {
+public abstract class Command {
 
-    void process(JsonNode payload, WhitelistProcessing whitelistProcessing);
+    private String commandRegex;
 
-    String getCommandRegex();
+    public abstract void process(JsonNode payload, WhitelistProcessing whitelistProcessing);
```
xstefank: maybe we should start adding some javadoc? here and/or on the class

src/test/java/org/xstefank/model/PersistentListTest.java
```diff
@@ -30,8 +30,8 @@ public void before() {
 
     @Test
     public void testIfUserIsSaved() {
-        testUserList.addUser(USERNAME);
-        Assert.assertTrue(testUserList.hasUsername(USERNAME));
+        testUserList.add(USERNAME);
```
xstefank: maybe it would be worthwhile to rename these tests to not reference userlist anymore?

src/main/java/org/xstefank/whitelist/Command.java
```diff
@@ -2,11 +2,29 @@
 
 import com.fasterxml.jackson.databind.JsonNode;
 
-public interface Command {
+/**
+ * Command for triggering CI
```
xstefank: nope, can be anything really

src/main/java/org/xstefank/whitelist/Command.java
```diff
@@ -2,11 +2,29 @@
...
-    void process(JsonNode payload, WhitelistProcessing whitelistProcessing);
+    private String commandRegex;
 
-    String getCommandRegex();
+    /**
+     * This method is used for command processing.
+     * You can define your own functionality of new command
+     * that you want to implement. Method is triggered automatically
+     * when the command regex is matched. You can set the regex
+     * in CommandsLoader class and use it in format.yaml.
```
xstefank: {@link CommandsLoader} ?

src/main/java/org/xstefank/whitelist/Command.java
```diff
@@ -2,11 +2,29 @@
...
+     * @param payload json received from github api.
+     * @param whitelistProcessing class that offers some useful functionality for your command.
```
xstefank: this can be more precise

src/test/java/org/xstefank/model/PersistentListTest.java
```diff
@@ -3,55 +3,162 @@
...
+    @Test
+    public void testSetElement() {
+        persistentList.add(FIRST);
+        int index = persistentList.indexOf(FIRST);
+        persistentList.set(index, SECOND);
+        FileLine line = TestUtils.getSpecificLine(persistentListFile, SECOND);
+
+        Assert.assertTrue(persistentList.contains(SECOND));
+        Assert.assertSame(index, persistentList.indexOf(SECOND));
+        Assert.assertSame(index + 1, line.getLineNumber());
```
xstefank: be careful with assertSame vs assertEquals

src/test/java/org/xstefank/model/PersistentListTest.java
```diff
@@ -3,55 +3,162 @@
...
     }
 
     @After
     public void after() {
-        TestUtils.deleteFileIfExists(userListFile);
+        TestUtils.deleteFileIfExists(persistentListFile);
+    }
+
+    private void assertsForAddAll(int indexOne, int indexTwo, int expectedOne, int expectedTwo) {
+        FileLine firstElLine = TestUtils.getSpecificLine(persistentListFile, FIRST);
```
xstefank: what does `El` stands for?

src/main/java/org/xstefank/model/PersistentList.java
```diff
@@ -30,32 +33,146 @@ public PersistentList(String dirName, String fileName) {
...
+    @Override
+    public String set(int index, String element) {
+        String oldValue = super.set(index, element);
+        rewriteFileContent(index);
+        return oldValue;
+    }
+
+    @Override
+    public boolean add(String s) {
+        if (contains(s) || !super.add(s)) {
```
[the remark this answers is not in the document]
lukasburda: I think i will remove that because of consistency (comment before).

src/main/java/org/xstefank/model/PersistentList.java
```diff
@@ -30,32 +33,146 @@ public PersistentList(String dirName, String fileName) {
...
+            return false;
+        }
+        try (FileWriter fw = new FileWriter(file, true)) {
+            fw.write(s + System.lineSeparator());
```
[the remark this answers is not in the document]
lukasburda: I agree up to a point.. If I call method `super.addAll(Collection)` it would not work because I will have to add line separator next to every single element in that collection.

src/main/java/org/xstefank/model/PersistentList.java
```diff
@@ -30,32 +33,146 @@ public PersistentList(String dirName, String fileName) {
...
+                randomAccessFile.writeBytes(lastProcessedElement + System.lineSeparator());
+            }
+        } catch (IOException e) {
+            if (lastProcessedElement == null) {
+                throw new IllegalStateException(e);
+            }
+            log.errorf("Cannot write element \"%s\" to list, Stacktrace: %s", lastProcessedElement, e);
+            int index = indexOf(lastProcessedElement);
+            super.remove(index);
+            rewriteFileContent(index);
```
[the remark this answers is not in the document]
lukasburda: I think it can't produce infinite loop.. Because I am removing every element that failed on previous line however, it could produce really long loop that will remove every single element from the list.

src/main/java/org/xstefank/model/PersistentList.java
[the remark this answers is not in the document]
xstefank: @lukasburda sorry I don't follow... How does that relate to extracting lineseparator to a field as I was saying in the original comment? You're code will be the same except for 
```suggestion
            fw.write(s + LINE_SEPARATOR);
```

src/main/java/org/xstefank/model/PersistentList.java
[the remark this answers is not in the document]
xstefank: @lukasburda I think you are removing element from parent list -- which is itself another error I missed so please take a look at that also -- index isn't changing as well as the content of the file so I still think that if some line is unable to be written you will cycle

src/test/java/org/xstefank/model/PersistentListTest.java
lukasburda: I though for numbers it is OK to use assertSame

xstefank: yes, they are, but be careful with for instance int == int and Integer == Integer

src/main/java/org/xstefank/model/PersistentList.java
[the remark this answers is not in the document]
lukasburda: Sorry I misunderstood.

src/main/java/org/xstefank/model/PersistentList.java
```diff
@@ -5,57 +5,157 @@
...
+                    super.add(line);
                 }
             }
         } catch (IOException e) {
-            throw new IllegalStateException("Cannot read file " + file.getPath(), e);
+            throw new IllegalStateException("Cannot access file " + file.getPath(), e);
+        }
+    }
+
+    private void tryToRewriteFileContent(int fromIndex) {
```
xstefank:
```suggestion
    private void rewriteFileContent(int fromIndex) {
```

src/main/java/org/xstefank/model/PersistentList.java
```diff
@@ -5,57 +5,157 @@
...
+            randomAccessFile.setLength(characterCount);
+            randomAccessFile.seek(characterCount);
+
+            for (int i = fromIndex; i < size(); i++) {
+                randomAccessFile.writeBytes(get(i) + Utils.LINE_SEPARATOR);
+            }
+        } catch (IOException e) {
+            log.error(e);
+            super.clear();
+            loadFileContent(file);
```
xstefank: If some element fails you clear whole parent list? and then load to it everything that have been successfully written to file?

take for instance: memory: [A, B, C], file: [A, B, C]
set(1, X);
memory: [A, X, C] file: start writing [A, X] then fail -> clear memory so: [] then load everything from file so:
memory: [A, X], file: [A, X]

----
so user called set(1, X) on list [A, B, C] and the result is [A, X] ? IMHO this is a really big error.

You should take care of alignment memory vs. file on the level of individual methods - even if this catch block would have made sense then clearing super list and loading it again in set() operation which concerns only one element is really redundant.

src/test/java/org/xstefank/FileLine.java
```diff
@@ -0,0 +1,24 @@
...
+    public FileLine(int lineNumber, String line) {
+        this.lineNumber = lineNumber;
+        this.content = line;
+    }
+
+    public boolean exists() {
+        return lineNumber >= 1 && content != null;
+    }
+
+    public int getLineNumber() {
```
xstefank: This doesn't seems to be used anywhere. What is the point there in having FileLine class at all?

Issue #42, opened 2019-02-22
Add control and test for token authentication


khermano:
method check in LatestCommitCheck throwing NullPointerException

xstefank:
@khermano when you report a bug it is usually a good practice to provide some background info (link to code, stacktrace, reproducer, environment, ...). Generally, the more you can help the developer who will be fixing this the better ;)

khermano:
Sorry. :)
So stacktrace is saying something like that: 
`ERROR [io.undertow.request] (default task-2) UT005023: Exception handling request to /tyr/pull-request: org.jboss.resteasy.spi.UnhandledException: java.lang.NullPointerException
	at org.jboss.resteasy.core.ExceptionHandler.handleApplicationException(ExceptionHandler.java:78)
	at org.jboss.resteasy.core.ExceptionHandler.handleException(ExceptionHandler.java:222)
	at org.jboss.resteasy.core.SynchronousDispatcher.writeException(SynchronousDispatcher.java:193)
	at org.jboss.resteasy.core.SynchronousDispatcher.invoke(SynchronousDispatcher.java:455)
	at org.jboss.resteasy.core.SynchronousDispatcher.lambda$invoke$4(SynchronousDispatcher.java:229)
	at org.jboss.resteasy.core.SynchronousDispatcher.lambda$preprocess$0(SynchronousDispatcher.java:135)
	at org.jboss.resteasy.core.interception.PreMatchContainerRequestContext.filter(PreMatchContainerRequestContext.java:355)
	at org.jboss.resteasy.core.SynchronousDispatcher.preprocess(SynchronousDispatcher.java:138)
	at org.jboss.resteasy.core.SynchronousDispatcher.invoke(SynchronousDispatcher.java:215)
	at org.jboss.resteasy.plugins.server.servlet.ServletContainerDispatcher.service(ServletContainerDispatcher.java:227)
	at org.jboss.resteasy.plugins.server.servlet.HttpServletDispatcher.service(HttpServletDispatcher.java:56)
	at org.jboss.resteasy.plugins.server.servlet.HttpServletDispatcher.service(HttpServletDispatcher.java:51)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:791)
	at io.undertow.servlet.handlers.ServletHandler.handleRequest(ServletHandler.java:74)
	at io.undertow.servlet.handlers.security.ServletSecurityRoleHandler.handleRequest(ServletSecurityRoleHandler.java:62)
[...]

PR #48, opened 2019-03-19
[Issue #42] Add control and test for token authentication

khermano:
Issue: https://github.com/xstefank/tyr/issues/42

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -14,7 +14,21 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <linkXRef>false</linkXRef>
         <version.com.fasterxml.jackson>2.9.8</version.com.fasterxml.jackson>
+        <version.wildfly>16.0.0.Final</version.wildfly>
+        <jboss.home>${project.build.directory}/wildfly-${version.wildfly}</jboss.home>
     </properties>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.jboss.arquillian</groupId>
+                <artifactId>arquillian-bom</artifactId>
+                <version>1.4.1.Final</version>
+                <scope>import</scope>
+                <type>pom</type>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
 
     <dependencies>
         <dependency>
@@ -25,9 +39,8 @@
 
         <dependency>
             <groupId>org.jboss.spec.javax.ws.rs</groupId>
-            <artifactId>jboss-jaxrs-api_2.0_spec</artifactId>
-            <version>1.0.1.Final</version>
-            <scope>compile</scope>
+            <artifactId>jboss-jaxrs-api_2.1_spec</artifactId>
+            <version>1.0.2.Final</version>
         </dependency>
 
         <dependency>
@@ -80,6 +93,18 @@
             <artifactId>jersey-client</artifactId>
             <version>2.22.2</version>
             <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.jboss.arquillian.junit</groupId>
+            <artifactId>arquillian-junit-container</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.wildfly</groupId>
+            <artifactId>wildfly-arquillian-container-managed</artifactId>
+            <version>8.2.1.Final</version>
         </dependency>
     </dependencies>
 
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -81,11 +92,27 @@
             <version>2.22.2</version>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.jboss.arquillian.junit</groupId>
+            <artifactId>arquillian-junit-container</artifactId>
+            <version>1.4.0.Final</version>
```
xstefank: cannot this version be taken from imported arquillian bom?

pom.xml
```diff
@@ -81,11 +92,27 @@
...
 
     <build>
         <finalName>tyr</finalName>
         <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.8.0</version>
+                <configuration>
+                    <source>1.8</source>
```
xstefank: you can do the same with <maven.compiler.source> and <maven.compiler.target> properties. Please move this configuration into properties

pom.xml
```diff
@@ -137,4 +164,85 @@
             </plugin>
         </plugins>
     </build>
+
+    <profiles>
+        <profile>
+            <id>wildfly-managed</id>
```
xstefank: maybe different name? what this really does is integration tests right?

pom.xml
```diff
@@ -137,4 +164,85 @@
...
+                <plugins>
+                    <plugin>
+                        <artifactId>maven-surefire-plugin</artifactId>
+                        <version>2.16</version>
+                        <configuration>
+                            <trimStackTrace>false</trimStackTrace>
+                            <systemPropertyVariables>
+                                <arquillian.launch>wildfly-managed</arquillian.launch>
+                                <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
+                                <jboss.home>${project.basedir}/target/wildfly-16.0.0.Final</jboss.home>
```
xstefank: please make the version of WFLY configurable in properties

pom.xml
```diff
@@ -137,4 +164,85 @@
...
+                                <goals>
+                                    <goal>integration-test</goal>
+                                    <goal>verify</goal>
+                                </goals>
+                                <configuration>
+                                    <systemPropertyVariables>
+                                        <arquillian.launch>wildfly-managed</arquillian.launch>
+                                        <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
+                                        <jboss.home>${project.basedir}/target/wildfly-16.0.0.Final</jboss.home>
+                                        <module.path>${project.basedir}/target/wildfly-16.0.0.Final/modules</module.path>
```
xstefank: this is same configuration as in surefire plugin. Is this necessary? If so why?

src/test/java/org/xstefank/api/AuthorizationTestIT.java
```diff
@@ -0,0 +1,39 @@
...
+import org.junit.runner.RunWith;
+import org.jboss.arquillian.junit.Arquillian;
+import org.xstefank.JaxRsApplication;
+
+import javax.ws.rs.core.UriBuilder;
+import java.net.URL;
+
+@RunWith(Arquillian.class)
+@RunAsClient
+public class AuthorizationTestIT {
```
xstefank: Is it authorization or authentication?

src/test/java/org/xstefank/api/AuthorizationTestIT.java
```diff
@@ -0,0 +1,39 @@
...
+    @ArquillianResource
+    private URL deploymentUrl;
+
+    @Test (expected=IllegalArgumentException.class)
+    public void testWrongAuthentication() {
+        GitHubAPI.getJsonFromUri(UriBuilder.fromPath(deploymentUrl.toString()).path("/test/fakeGithub_ReturnStatusUnauthorized").build());
+    }
+
+    @Test (expected=IllegalArgumentException.class)
+    public void testWrongAuthentication2() {
```
xstefank: please rename this

src/test/java/org/xstefank/api/AuthorizationTestIT.java
```diff
@@ -0,0 +1,39 @@
...
+                .create(WebArchive.class, "test.war")
+                .addClasses(FakeGitHub.class, JaxRsApplication.class);
+    }
+
+    @ArquillianResource
+    private URL deploymentUrl;
+
+    @Test (expected=IllegalArgumentException.class)
+    public void testWrongAuthentication() {
+        GitHubAPI.getJsonFromUri(UriBuilder.fromPath(deploymentUrl.toString()).path("/test/fakeGithub_ReturnStatusUnauthorized").build());
```
xstefank: there is also fromURI method in UriBuilder

src/test/java/org/xstefank/api/FakeGitHub.java
```diff
@@ -0,0 +1,21 @@
+package org.xstefank.api;
+
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.core.Response;
+
+@Path("/fakeGitHub_")
```
xstefank: just curiosity -- why underscore at the end?

pom.xml
khermano: unfortunately no... at least what I know...http://arquillian.org/modules/core-platform/#artifacts

pom.xml
```diff
@@ -81,11 +93,28 @@
...
     </dependencies>
 
     <build>
         <finalName>tyr</finalName>
         <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.8.0</version>
+            </plugin>
```
xstefank: is compiler plugin definition still required if there is no additional configuration?

pom.xml
```diff
@@ -137,4 +166,65 @@
...
+                                <goals>
+                                    <goal>integration-test</goal>
+                                    <goal>verify</goal>
+                                </goals>
+                                <configuration>
+                                    <systemPropertyVariables>
+                                        <arquillian.launch>wildfly-managed</arquillian.launch>
+                                        <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
+                                        <jboss.home>${project.basedir}/target/wildfly-${version.wildfly}</jboss.home>
+                                        <module.path>${project.basedir}/target/wildfly-${version.wildfly}/modules</module.path>
```
xstefank: there is also project.build.directory

xstefank: and I believe you can reuse the path defined on the previous line, if not directly than you can define it as global property

src/test/java/org/xstefank/api/AuthenticationTestIT.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    @Test (expected=IllegalArgumentException.class)
+    public void testAuthenticationStatusInternalServerError() {
+        Assert.assertEquals(Response.Status.INTERNAL_SERVER_ERROR.getStatusCode(), getEndPointResponse(RETURN_STATUS_INTERNAL_SERVER_ERROR_PATH).getStatus());
+        testAuthentication(RETURN_STATUS_INTERNAL_SERVER_ERROR_PATH);
+    }
+
+    private Response getEndPointResponse(String endpointPath) {
+        Client restClient = ClientBuilder.newClient();
```
xstefank: client instance can be reused, and should be closed after tests - try before/after class annotations

src/test/java/org/xstefank/api/AuthenticationTestIT.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Test (expected=IllegalArgumentException.class)
+    public void testAuthenticationStatusUnauthorized() {
+        Assert.assertEquals(Response.Status.UNAUTHORIZED.getStatusCode(), getEndPointResponse(RETURN_STATUS_UNAUTHORIZED_PATH).getStatus());
+        testAuthentication(RETURN_STATUS_UNAUTHORIZED_PATH);
+    }
+
+    @Test (expected=IllegalArgumentException.class)
+    public void testAuthenticationStatusInternalServerError() {
+        Assert.assertEquals(Response.Status.INTERNAL_SERVER_ERROR.getStatusCode(), getEndPointResponse(RETURN_STATUS_INTERNAL_SERVER_ERROR_PATH).getStatus());
```
xstefank: I'm not sure if this is the right approach as you are basically making same rest call twice if I'm not mistaken. But ok, if you want to verify that the paths are right I will merge this

khermano: Thanks :)

PR #53, opened 2019-03-22
Issue #52: Commit check should take into account only first commit me…

### LINKED ISSUE #52: Commit check should take into account only first commit message
if you include multiple messages in the commit they are separated by end of line characters. We should take only the first line if applicable.

xstefank:
…ssage

Issue: https://github.com/xstefank/tyr/issues/52

resolves #52


xstefank:
@khermano @lukasburda can you review please?

src/test/resources/json/multipleCommitMessagesPayload.json
```diff
@@ -0,0 +1,7 @@
+[
+  {
+    "commit": {
+      "message": "Test commit\r\nSecond commit message"
```
khermano: not sure but whould not be "\n" enough?

xstefank: because github is returnig \r\n

Issue #38, opened 2018-12-31
Remove JBOSS_CONFIG_DIR and make config directory configurable by system property


xstefank:

lukasburda:
This can be closed.

PR #58, opened 2019-03-26
[Issue #38] Remove JBOSS_CONFIG_DIR and make config directory configurable by system property

lukasburda:
Issue: https://github.com/xstefank/tyr/issues/38

### DIFF
--- a/src/main/java/org/xstefank/model/Utils.java
+++ b/src/main/java/org/xstefank/model/Utils.java
@@ -24,7 +24,7 @@ public class Utils {
     public static final String CONFIG_FILE = "config.properties";
     public static final String TOKEN_PROPERTY = "github.oauth.token";
     public static final String TOKEN_ENV = "GITHUB_OAUTH_TOKEN";
-    public static final String JBOSS_CONFIG_DIR = "jboss.server.config.dir";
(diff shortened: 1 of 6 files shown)


xstefank:
@lukasburda CI is failing

src/main/java/org/xstefank/model/Utils.java
```diff
@@ -62,4 +63,12 @@ public static URL getFormatUrl() throws MalformedURLException {
             target = TyrProperties.getProperty(TEMPLATE_FORMAT_URL);
         return target != null ? new URL(target) : null;
     }
+
+    private static String getConfigDirectory() {
+        String path = System.getProperty(TYR_CONFIG_DIR);
+        if (path == null) {
+            throw new IllegalStateException("Configuration directory property is not set");
```
xstefank: I just wonder if this needs to be required, can we take for default current directory for instance?

lukasburda: I think that we never know which application server would be used and I think every server has different property. What comes to my mind is that we could have maybe 3-4 different application servers properties that we could try to parse and if all of them will be null and also that `TYR_CONFIG_DIR` will be null it could throw an exception. What do you think about that?

Issue #57, opened 2019-03-25
Implement Team City retest failed functionality


xstefank:

xstefank:
PR merged.

PR #87, opened 2019-08-05
[Issue #57] Implement Team City retest failed functionality

khermano:
Issue: https://github.com/jboss/tyr/issues/57

### DIFF
--- a/tyr-runner/src/main/java/org/jboss/tyr/ci/ContinuousIntegration.java
+++ b/tyr-runner/src/main/java/org/jboss/tyr/ci/ContinuousIntegration.java
@@ -15,13 +15,14 @@
  */
 package org.jboss.tyr.ci;
 
+import org.jboss.tyr.InvalidPayloadException;
 import javax.json.JsonObject;
 
 public interface ContinuousIntegration {
 
-    void triggerBuild(JsonObject prPayload);
+    void triggerBuild(JsonObject prPayload) throws InvalidPayloadException;
 
-    void triggerFailedBuild(JsonObject prPayload);
+    void triggerFailedBuild(JsonObject prPayload) throws InvalidPayloadException;
 
     void init();
 }
--- a/tyr-runner/src/main/java/org/jboss/tyr/ci/TeamCityCI.java
+++ b/tyr-runner/src/main/java/org/jboss/tyr/ci/TeamCityCI.java
@@ -16,15 +16,21 @@
 package org.jboss.tyr.ci;
 
 import org.jboss.logging.Logger;
+import org.jboss.tyr.InvalidPayloadException;
 import org.jboss.tyr.model.TyrProperties;
 import org.jboss.tyr.model.Utils;
 import org.jboss.tyr.model.json.BuildJson;
-
+import org.jboss.tyr.model.json.Property;
+import org.jboss.tyr.model.json.SnapshotDependencies;
+import org.jboss.tyr.model.json.SnapshotDependency;
 import javax.json.JsonObject;
-import javax.ws.rs.client.Client;
-import javax.ws.rs.client.ClientBuilder;
-import javax.ws.rs.client.Entity;
+import javax.json.bind.Jsonb;
+import javax.json.bind.JsonbBuilder;
 import javax.ws.rs.client.WebTarget;
+import javax.ws.rs.client.Entity;
+import javax.ws.rs.client.Client;
+import javax.ws.rs.client.Invocation;
+import javax.ws.rs.client.ClientBuilder;
 import javax.ws.rs.core.HttpHeaders;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
@@ -38,18 +44,24 @@ import java.util.Map;
 public class TeamCityCI implements ContinuousIntegration {
 
     public static final String NAME = "TeamCity";
-
     public static final String HOST_PROPERTY = "teamcity.host";
     public static final String PORT_PROPERTY = "teamcity.port";
     public static final String USER_PROPERTY = "teamcity.user";
     public static final String PASSWORD_PROPERTY = "teamcity.password";
     public static final String BUILD_CONFIG = "teamcity.branch.mapping";
+    public static final String BUILD_PATH = "/app/rest/buildQueue";
+    public static final String SNAPSHOTDEPENDENCIES_PATH_PREFIX = "app/rest/buildTypes/id:";
+    public static final String SNAPSHOTDEPENDENCIES_PATH_SUFFIX = "/snapshot-dependencies";
+    public static final String BUILD_WITH_SAME_REVISIONS_PROPERTY = "take-started-build-with-same-revisions";
+    public static final String SUCCESSFUL_BUILDS_ONLY_PROPERTY = "take-successful-builds-only";
 
     private static final Logger log = Logger.getLogger(TeamCityCI.class);
 
     private String baseUrl;
     private String encryptedCredentials;
     private Map<String, String> branchMappings;
+    private Client client;
+    private Jsonb jsonb;
 
     @Override
     public void init() {
@@ -60,47 +72,16 @@ public class TeamCityCI implements ContinuousIntegration {
                 TyrProperties.getProperty(PASSWORD_PROPERTY));
 
         this.branchMappings = parseBranchMapping(TyrProperties.getProperty(BUILD_CONFIG));
+
+        this.client = ClientBuilder.newClient();
+
+        this.jsonb = JsonbBuilder.create();
     }
 
     @Override
-    public void triggerBuild(JsonObject prPayload) {
-        String branch = prPayload.getJsonObject(Utils.BASE).getString(Utils.REF);
-        if (!branchMappings.containsKey(branch)) {
-            return;
-        }
-
-        int pull = prPayload.getInt(Utils.NUMBER);
-        String sha = prPayload.getJsonObject(Utils.HEAD).getString(Utils.SHA);
-        String buildId = branchMappings.get(branch);
-
-        Client client = ClientBuilder.newClient();
-        URI statusUri = UriBuilder
-                .fromUri(baseUrl)
-                .path("/app/rest/buildQueue")
-                .build();
-
-        WebTarget target = client.target(statusUri);
-        Entity<BuildJson> json = Entity.json(new BuildJson("pull/" + pull, buildId, sha, pull, branch));
-
-        Response response = null;
-
-        try {
-            response = target.request()
-                .header(HttpHeaders.ACCEPT_ENCODING, "UTF-8")
-                .header(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON)
-                .header(HttpHeaders.AUTHORIZATION, "Basic " + encryptedCredentials)
-                .post(json);
-
-            log.info("Teamcity status update: " + response.getStatus());
-        } catch (Throwable e) {
-            log.error("Cannot run Team city build", e);
-        } finally {
-            if (response != null) {
-                response.close();
-            }
-        }
+    public void triggerBuild(JsonObject prPayload) throws InvalidPayloadException {
+        addJsonToBuildQueue(prPayload, false);
     }
-
 
     private String encryptCredentials(String username, String password) {
         String authStr = username + ":" + password;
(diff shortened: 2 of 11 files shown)


tyr-runner/src/main/java/org/jboss/tyr/ci/ContinuousIntegration.java
```diff
@@ -19,9 +19,9 @@
 
 public interface ContinuousIntegration {
 
-    void triggerBuild(JsonObject prPayload);
+    void triggerRetesting(JsonObject prPayload);
```
xstefank: what is wrong with the original?

tyr-runner/src/main/java/org/jboss/tyr/ci/ContinuousIntegration.java
```diff
@@ -19,9 +19,9 @@
...
 
-    void triggerFailedBuild(JsonObject prPayload);
+    void triggerRetestingFailed(JsonObject prPayload);
```
xstefank:
```suggestion
    void triggerFailedTests(JsonObject prPayload);
```

tyr-runner/src/main/java/org/jboss/tyr/ci/TeamCityCI.java
```diff
@@ -101,29 +124,100 @@ public void triggerBuild(JsonObject prPayload) {
...
-    private String encryptCredentials(String username, String password) {
-        String authStr = username + ":" + password;
-        return DatatypeConverter.printBase64Binary(authStr.getBytes(StandardCharsets.UTF_8));
+    @Override
+    public void triggerRetestingFailed(JsonObject prPayload) {
+        setTriggerBuildProperties(prPayload);
+        if (!branchMappings.containsKey(branch)) {
+            return;
+        }
+        setSnapshotDependencies(buildId, true);
```
xstefank: Can you refactor this somehow please? This method and triggerRetesting differ only in the boolean value passed to this invocation.

tyr-runner/src/main/java/org/jboss/tyr/ci/TeamCityCI.java
```diff
@@ -101,29 +124,100 @@ public void triggerBuild(JsonObject prPayload) {
...
 
-        for (String mapping : mappings.split(",")) {
-            String[] parts = mapping.split("=>");
-            branchMap.put(parts[0].trim(), parts[1].trim());
+        try {
+            response = target.request()
+                .header(HttpHeaders.ACCEPT_ENCODING, "UTF-8")
+                .header(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON)
+                .header(HttpHeaders.AUTHORIZATION, "Basic " + encryptedCredentials)
+                .header(HttpHeaders.ACCEPT, MediaType.APPLICATION_JSON)
```
xstefank: maybe you can return somewhere this whole request builder object so you don't need to repeat all the headers everytime?

xstefank: or even more generic `getJson(path)`

tyr-runner/src/main/java/org/jboss/tyr/ci/TeamCityCI.java
```diff
@@ -101,29 +124,100 @@ public void triggerBuild(JsonObject prPayload) {
...
         }
+        return snapshotDependencies;
+    }
 
-        return branchMap;
+    private void setTriggerBuildProperties(JsonObject prPayload) {
+        branch = prPayload.getJsonObject(Utils.BASE).getString(Utils.REF);
+        pull = prPayload.getInt(Utils.NUMBER);
+        sha = prPayload.getJsonObject(Utils.HEAD).getString(Utils.SHA);
+        buildId = branchMappings.get(branch);
```
xstefank: so you are setting these fields on every invocation of triggerRetesting*. This kind of storing values is inefficient. If they are stored only for one method invocation then it is not necessary to store them in fields.

tyr-runner/src/main/java/org/jboss/tyr/ci/TeamCityCI.java
```diff
@@ -101,29 +124,100 @@ public void triggerBuild(JsonObject prPayload) {
...
     }
 
-    @Override
-    public void triggerFailedBuild(JsonObject prPayload) {
-        throw new UnsupportedOperationException("Method is not implemented");
+    private WebTarget returnTeamCityTarget(String path) {
```
xstefank:
```suggestion
    private WebTarget getTeamCityTarget(String path) {
```

tyr-runner/src/main/java/org/jboss/tyr/ci/TeamCityCI.java
```diff
@@ -101,29 +124,100 @@ public void triggerBuild(JsonObject prPayload) {
...
+        Client client = ClientBuilder.newClient();
```
xstefank: client object can be reused

tyr-runner/src/main/java/org/jboss/tyr/model/json/Property.java
```diff
@@ -15,11 +15,18 @@
  */
 package org.jboss.tyr.model.json;
 
+import javax.json.bind.annotation.JsonbPropertyOrder;
+
+@JsonbPropertyOrder({"name", "value"})
```
xstefank: why is this important, please?

tyr-runner/src/main/java/org/jboss/tyr/model/json/SnapshotDependencies.java
```diff
@@ -0,0 +1,20 @@
+package org.jboss.tyr.model.json;
+
+import javax.json.bind.annotation.JsonbProperty;
+import javax.json.bind.annotation.JsonbPropertyOrder;
+import java.math.BigDecimal;
+import java.util.List;
+
+@JsonbPropertyOrder({"count", "snapshot-dependency"})
```
xstefank: why is this important, please?

tyr-runner/src/main/java/org/jboss/tyr/model/json/SnapshotDependency.java
```diff
@@ -0,0 +1,19 @@
+package org.jboss.tyr.model.json;
+
+import javax.json.bind.annotation.JsonbProperty;
+import javax.json.bind.annotation.JsonbPropertyOrder;
+
+@JsonbPropertyOrder({"id", "type", "properties", "source-buildType"})
```
xstefank: sorry but again why is the order important?

tyr-runner/src/main/java/org/jboss/tyr/ci/ContinuousIntegration.java
khermano: There still is the original method https://github.com/khermano/tyr/commit/0bbcf1a3ad81bba448f8b38d878b73449c539915#diff-2495a7c9254abaaeb22b96823ce7bb5aR103 .... I just want to separate its functionality because of DRY principle as I use this method in both triggerRetesting() and triggerRetestingFailed() methods.  :)

tyr-runner/src/main/java/org/jboss/tyr/ci/TeamCityCI.java
khermano: Yes, of course. :)

tyr-runner/src/main/java/org/jboss/tyr/ci/ContinuousIntegration.java
xstefank: that is a private method in the implementation. That's not a reason to rename interface.

khermano: Sure, I understand your point. I will rename it back and change name of the private method. :)

tyr-runner/src/main/java/org/jboss/tyr/model/json/Property.java
khermano: It is not anymore...there was some problem with mapping but that is irrelevant now...I will remove it. Sorry, my bad. :) This applies to all @JsonbPropertyOrder annotation which I added.

khermano:
Messages are flowing between Tyr and TeamCity:

- Tyr receives PR comment JSON
- trigger* method collects necessary info from upper JSON and config.properties file
- Tyr invokes e.g. `http://localhost:8111/app/rest/buildTypes/id:TestRepo_Aggregator/snapshot-dependencies` to get JSON with snapshot dependencies (contains build configurations from build chain) of aggregation build configuration (buildId = TestRepo_Aggregator)
- Tyr put same JSON with snapshot dependencies on the same URL changing only `take-started-build-with-same-revisions` and `take-successful-builds-only` properties in every snapshot dependency depending on whether we want run all builds or only failed ones (true if we want to test only failed ones)
- Tyr triggers build by posting build JSON to e.g. `http://localhost:8111/app/rest/buildQueue`

tyr-runner/src/main/java/org/jboss/tyr/ci/TeamCityCI.java
```diff
@@ -73,57 +110,103 @@ public void triggerBuild(JsonObject prPayload) {
...
-    }
+        if (snapshotDependencies.snapshotDependencyCount.intValue() < 1) {
+            log.info("No snapshot dependency.");
+            return;
+        }
 
-    private String getBaseUrl(String host, int port) {
-        return String.format("http://%s:%d/httpAuth", host, port);
+        for (SnapshotDependency dependency: snapshotDependencies.snapshotDependencies) {
+            for (Property prop :dependency.properties.propertyList) {
```
xstefank: just minor code style preference, use space around ':' on both sides in for-each loops please

tyr-runner/src/main/java/org/jboss/tyr/ci/TeamCityCI.java
```diff
@@ -73,57 +110,103 @@ public void triggerBuild(JsonObject prPayload) {
...
+                if (prop.name.equals("take-started-build-with-same-revisions") || prop.name.equals("take-successful-builds-only")) {
+                    prop.value = String.valueOf(retestOnlyFailed);
+                }
+            }
+        }
+
+        Jsonb jsonb = JsonbBuilder.create();
+        String stringJsonb = jsonb.toJson(snapshotDependencies);
```
xstefank: it's not a string with jsonb right? it's string with snapshotDependencies :)

tyr-runner/src/main/java/org/jboss/tyr/ci/TeamCityCI.java
```diff
@@ -73,57 +110,103 @@ public void triggerBuild(JsonObject prPayload) {
...
+        Entity<String> json = Entity.json(stringJsonb);
+
+        String path = "app/rest/buildTypes/id:" + buildId + "/snapshot-dependencies";
+        WebTarget target = getTeamCityTarget(path);
+        Response response = null;
+
+        try {
+            response = getHeaderBuilder(target).put(json);
+            log.info("Putting JSON with snapshot dependencies in Teamcity. Status: " + response.getStatus());
```
xstefank: Sending JSON ...

tyr-runner/src/main/java/org/jboss/tyr/ci/TeamCityCI.java
```diff
@@ -73,57 +110,103 @@ public void triggerBuild(JsonObject prPayload) {
...
+        WebTarget target = getTeamCityTarget(path);
+        Response response = null;
+        SnapshotDependencies snapshotDependencies;
 
-        for (String mapping : mappings.split(",")) {
-            String[] parts = mapping.split("=>");
-            branchMap.put(parts[0].trim(), parts[1].trim());
+        try {
+            response = getHeaderBuilder(target).get();
+            log.info("Getting JSON with snapshot dependencies from Teamcity. Status: " + response.getStatus());
```
xstefank: use debug for these log messages please (also putting... above)

tyr-runner/src/main/java/org/jboss/tyr/ci/TeamCityCI.java
```diff
@@ -73,57 +110,103 @@ public void triggerBuild(JsonObject prPayload) {
...
+
+            String json = response.readEntity(String.class);
+            Jsonb jsonb = JsonbBuilder.create();
```
xstefank: you can cache Jsonb object in a field

tyr-runner/src/main/java/org/jboss/tyr/ci/TeamCityCI.java
```diff
@@ -73,57 +110,103 @@ public void triggerBuild(JsonObject prPayload) {
...
+        } finally {
+            if (response != null) {
+                response.close();
+            }
         }
+        return snapshotDependencies;
+    }
 
-        return branchMap;
+    private Invocation.Builder getHeaderBuilder(WebTarget target) {
```
xstefank:
```suggestion
    private Invocation.Builder getRequestBuilder(WebTarget target) {
```

tyr-runner/src/main/java/org/jboss/tyr/model/json/Property.java
```diff
@@ -20,6 +20,10 @@
     public String name;
     public Object value;
 
+    public Property() {
+
+    }
```
xstefank: why is this needed?

tyr-runner/src/main/java/org/jboss/tyr/model/json/SnapshotDependencies.java
```diff
@@ -0,0 +1,18 @@
...
+
+    @JsonbProperty("count")
+    public BigDecimal snapshotDependencyCount;
+
+    @JsonbProperty("snapshot-dependency")
+    public List<SnapshotDependency> snapshotDependencies;
+
+    public SnapshotDependencies() {
+
+    }
```
xstefank: also, why is the default constructor not enough? Also in other Json classes. Did you try to run the app without these empty constructors?

tyr-runner/src/test/java/org/jboss/tyr/ci/TeamCityCITest.java
```diff
@@ -15,12 +15,55 @@
...
+    @Test
+    public void testTriggerFailedTestsNullParameters() {
+        exception.expect(NullPointerException.class);
+        new TeamCityCI().triggerFailedTests(null);
+    }
+
+    @Test
+    public void testTriggerBuildPrPayloadMissingPullRequestParameter() {
+        exception.expect(NullPointerException.class);
+        new TeamCityCI().triggerBuild(TestUtils.EMPTY_PAYLOAD);
```
xstefank: shouldn't this raise something like "Invalid payload exception"? Also for other tests after this one.

tyr-runner/src/main/java/org/jboss/tyr/model/json/Property.java
khermano: cause I am reusing this already existing class for jsonb unmarshalling and it requires noargconstructor :) when I remove it I am getting NPE here https://github.com/khermano/tyr/commit/9fe776e81c5361c482c3abf7bda3e18fc90b6b5e#diff-2495a7c9254abaaeb22b96823ce7bb5aR183

tyr-runner/src/main/java/org/jboss/tyr/model/json/SnapshotDependencies.java
khermano: there you are right :) these I don't need because of default constructor...sorry for that, they are only remains when I was trying to use parametric constructor too..removing them :)

tyr-runner/src/main/java/org/jboss/tyr/ci/TeamCityCI.java
```diff
@@ -60,37 +67,69 @@ public void init() {
...
+            pull = prPayload.getInt(Utils.NUMBER);
+            sha = prPayload.getJsonObject(Utils.HEAD).getString(Utils.SHA);
+            buildId = branchMappings.get(branch);
+        } catch (NullPointerException e) {
+            throw new InvalidPayloadException("Invalid payload, can't retrieve all elements. Message was: " + e.getMessage());
+        }
+
+        setSnapshotDependencies(buildId, retestOnlyFailed);
+
+        String path = "/app/rest/buildQueue";
```
xstefank: this could be a constant for easier access, but not critical

tyr-runner/src/main/java/org/jboss/tyr/ci/TeamCityCI.java
```diff
@@ -101,29 +140,81 @@ public void triggerBuild(JsonObject prPayload) {
...
+        if (snapshotDependencies.snapshotDependencyCount.intValue() < 1) {
+            log.info("No snapshot dependency.");
+            return;
+        }
 
-    private String getBaseUrl(String host, int port) {
-        return String.format("http://%s:%d/httpAuth", host, port);
+        for (SnapshotDependency dependency : snapshotDependencies.snapshotDependencies) {
+            for (Property prop : dependency.properties.propertyList) {
+                if (prop.name.equals("take-started-build-with-same-revisions") || prop.name.equals("take-successful-builds-only")) {
```
xstefank: also better in constants

tyr-runner/src/main/java/org/jboss/tyr/ci/TeamCityCI.java
```diff
@@ -101,29 +140,81 @@ public void triggerBuild(JsonObject prPayload) {
...
+                    prop.value = String.valueOf(retestOnlyFailed);
+                }
+            }
+        }
+
+        Jsonb jsonb = JsonbBuilder.create();
```
xstefank: better to be cached. No need to create a new instance on every method call.

tyr-runner/src/main/java/org/jboss/tyr/ci/TeamCityCI.java
```diff
@@ -101,29 +140,81 @@ public void triggerBuild(JsonObject prPayload) {
...
+
+        String path = "app/rest/buildTypes/id:" + buildId + "/snapshot-dependencies";
+        WebTarget target = getTeamCityTarget(path);
+        Response response = null;
+
+        try {
+            response = getRequestBuilder(target).put(json);
+            log.debug("Sending JSON with snapshot dependencies in Teamcity. Status: " + response.getStatus());
+        } catch (Throwable e) {
+            log.error("Cannot retrieve JSON from " + target.getUri(), e);
```
xstefank: you are sending JSON right? :)

tyr-runner/src/main/java/org/jboss/tyr/ci/TeamCityCI.java
```diff
@@ -101,29 +140,81 @@ public void triggerBuild(JsonObject prPayload) {
...
+            if (response != null) {
+                response.close();
+            }
+        }
     }
 
-    private Map<String, String> parseBranchMapping(String mappings) {
-        Map<String, String> branchMap = new HashMap<>();
+    private SnapshotDependencies getSnapshotDependencies(String buildId) {
+        String path = "app/rest/buildTypes/id:" + buildId + "/snapshot-dependencies";
```
xstefank: this is even duplicated, so please move it to separate method

tyr-runner/src/main/java/org/jboss/tyr/ci/TeamCityCI.java
```diff
@@ -101,29 +140,81 @@ public void triggerBuild(JsonObject prPayload) {
...
 
-        for (String mapping : mappings.split(",")) {
-            String[] parts = mapping.split("=>");
-            branchMap.put(parts[0].trim(), parts[1].trim());
+        try {
+            response = getRequestBuilder(target).get();
+            log.debug("Getting JSON with snapshot dependencies from Teamcity. Status: " + response.getStatus());
+
+            String json = response.readEntity(String.class);
+            snapshotDependencies = JsonbBuilder.create().fromJson(json, SnapshotDependencies.class);
```
xstefank: and new instance of Jsonb again

khermano:
> @khermano looks good. Two additional points to what I've added inline:
> 
> 1/ Each file must have a license header. Check other existing classes.
> 2/ Does this require some special setup of TeamCity? If yes, do we have it documented?

About point number 2: https://github.com/khermano/tyr/blob/master/tyr-runner/src/test/running_teamcity_test_instance.adoc ...this is all we have for now

tyr-runner/src/main/java/org/jboss/tyr/ci/TeamCityCI.java
```diff
@@ -101,29 +146,80 @@ public void triggerBuild(JsonObject prPayload) {
...
+            }
+        }
+
+        String snapshotDependenciesString = jsonb.toJson(snapshotDependencies);
+        Entity<String> json = Entity.json(snapshotDependenciesString);
+        Response response = null;
+
+        try {
+            response = getSnapshotDependenciesRequestBuilder(buildId).put(json);
+            log.debug("Sending JSON with snapshot dependencies in Teamcity. Status: " + response.getStatus());
```
xstefank:
```suggestion
            log.debug("Sent JSON with snapshot dependencies to Teamcity. Status: " + response.getStatus());
```

tyr-runner/src/main/java/org/jboss/tyr/ci/TeamCityCI.java
```diff
@@ -101,29 +146,80 @@ public void triggerBuild(JsonObject prPayload) {
...
+    private SnapshotDependencies getSnapshotDependencies(String buildId) {
+        Response response = null;
+        SnapshotDependencies snapshotDependencies;
 
-        for (String mapping : mappings.split(",")) {
-            String[] parts = mapping.split("=>");
-            branchMap.put(parts[0].trim(), parts[1].trim());
+        try {
+            response = getSnapshotDependenciesRequestBuilder(buildId).get();
+            log.debug("Getting JSON with snapshot dependencies from Teamcity. Status: " + response.getStatus());
```
xstefank:
```suggestion
            log.debug("Got JSON with snapshot dependencies from Teamcity. Status: " + response.getStatus());
```

xstefank:
@khermano thanks!

khermano:
np :) thank you

Issue #97, opened 2020-02-05
Add usage of InvalidPayloadException


khermano:
There are couple of places in the code where the InvalidPayloadException would be more suitable than NullPointerException

PR #98, opened 2020-02-13
[Issue #97] Add usage of InvalidPayloadException

khermano:
Issue: https://github.com/jboss/tyr/issues/97

### DIFF
--- a/tyr-runner/src/main/java/org/jboss/tyr/api/GitHubAPI.java
+++ b/tyr-runner/src/main/java/org/jboss/tyr/api/GitHubAPI.java
@@ -16,6 +16,7 @@
 package org.jboss.tyr.api;
 
 import org.jboss.logging.Logger;
+import org.jboss.tyr.InvalidPayloadException;
 import org.jboss.tyr.model.CommitStatus;
 import org.jboss.tyr.model.StatusPayload;
 import org.jboss.tyr.model.TyrProperties;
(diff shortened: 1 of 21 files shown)


tyr-runner/src/main/java/org/jboss/tyr/api/GitHubAPI.java
```diff
@@ -119,14 +120,22 @@ static JsonReader getJSONReader(URI uri) {
...
 
-    private static URI getCommitsUri(JsonObject prPayload) {
-        String url = prPayload.getJsonObject(Utils.PULL_REQUEST).getString(Utils.COMMITS_URL);
-        return URI.create(url);
+    private static URI getCommitsUri(JsonObject prPayload) throws InvalidPayloadException {
+        try {
+            String url = prPayload.getJsonObject(Utils.PULL_REQUEST).getString(Utils.COMMITS_URL);
+            return URI.create(url);
+        } catch (NullPointerException e) {
+            throw new InvalidPayloadException("Invalid payload, can't retrieve URL. Message was: " + e.getMessage());
```
xstefank: wouldn't it be better to add NPE as the cause of IPE?

khermano: sure...something like that: "Failed with NullPointerException, can't retrieve URL. Message was: .." ?

xstefank:
Thanks @khermano

Issue #44, opened 2019-03-06
Move configuration to MP Config


xstefank:
at least:

- location of format.yaml
- github token
- whitelist enabled

And then anything else that may make sense.

lukasburda:
Can I start working on this?

xstefank:
@lukasburda sure, thanks

xstefank:
Resolved.

PR #127, opened 2020-04-06
[Issue #44] Move configuration to MP Config

lukasburda:
Issue: https://github.com/jboss/tyr/issues/44


tyr-runner/src/main/java/org/jboss/tyr/ci/TeamCityProperties.java
```diff
@@ -0,0 +1,76 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.tyr.ci;
+
+import io.quarkus.arc.config.ConfigProperties;
+import java.util.Optional;
+
+@ConfigProperties(prefix = "teamcity")
+public class TeamCityProperties {
```
xstefank, line 22: Can we use interface here? Only for consistency.

tyr-runner/src/test/java/org/jboss/tyr/command/CommandTest.java
```diff
@@ -34,23 +34,23 @@
...
     WhitelistProcessing whitelistProcessing;
 
+    @Inject
+    TestCI testCI;
+
     @BeforeAll
     public static void beforeAll() {
-        userListFile = new File(System.getProperty(Utils.TYR_CONFIG_DIR), Utils.USERLIST_FILE_NAME);
-        adminListFile = new File(System.getProperty(Utils.TYR_CONFIG_DIR), Utils.ADMINLIST_FILE_NAME);
+        userListFile = new File(TestUtils.TARGET_DIR, Utils.USERLIST_FILE_NAME);
```
xstefank, line 48: I am missing probably where this is picked up? The new name of the property is tyr.users.dir right? If not set the current directory is used which can be /target but it would be better to set the property to not rely on it.

lukasburda: It is picked up from application.properties file [https://github.com/jboss/tyr/blob/6bd501d52c268c11d2e7022902f99cf15e02ebb6/tyr-runner/src/test/resources/application.properties#L2](url)

tyr-runner/src/main/java/org/jboss/tyr/ci/CILoader.java
```diff
@@ -15,22 +15,20 @@
...
-    }
-
-    public static ContinuousIntegration getCI(String key) {
-        return CIs.get(key);
-    }
-
-    static void addCI(String key, ContinuousIntegration continuousIntegration) {
-        CIs.put(key, continuousIntegration);
+    public Optional<ContinuousIntegration> getCI(String key) {
+        return CIs.stream().filter(ci -> ci.getClass().getSimpleName().contains(key)).findFirst();
```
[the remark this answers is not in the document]
lukasburda: So should I resolve this now?

tyr-runner/src/main/java/org/jboss/tyr/model/TyrConfiguration.java
```diff
@@ -17,4 +17,18 @@
...
     boolean whitelistEnabled();
 
+    @ConfigProperty(name = "github.oauth.token")
+    String oauthToken();
+
+    @ConfigProperty(name = "users.dir")
+    Optional<String> whitelistDir();
+
+    @ConfigProperties(prefix = "teamcity")
+    interface TeamCityProperties {
```
xstefank, line 27: I personally don't like this design. This is "outside" package private interface so it should be in a separate file. Just make the prefix = "tyr.teamcity"

PR #139, opened 2021-01-11
[issue #37] Add configuration for optional checks

### LINKED ISSUE #37: Add configuration for optional checks
Add possibility to have optional checks / rows - if present they should be checked otherwise nothing happens

petrberan:
Resolves #37


tyr-runner/src/main/java/org/jboss/tyr/check/DescriptionCheck.java
```diff
@@ -48,6 +52,19 @@ public String check(JsonObject payload) {
...
                     }
                 }
+                for (OptionalPatterns optionalRow : optionalRows){
+                    Matcher preconditionMatcher = optionalRow.getPrecondition().matcher(line);
+                    if (preconditionMatcher.matches()){
+                        Matcher patternMatcher = optionalRow.getPattern().matcher(line);
+                        if (!patternMatcher.matches()){
+                            return optionalRow.getMessage();
+                        }
+                        if (patternMatcher.matches()){
```
xstefank, line 62: maybe you `else` so you don't need to repeat the condition computation.

tyr-runner/src/main/java/org/jboss/tyr/check/DescriptionCheck.java
```diff
@@ -25,21 +26,24 @@
...
 
     @Override
     public String check(JsonObject payload) {
         List<RegexDefinition> requiredRows = new ArrayList<>(rows);
+        List<OptionalPatterns> optionalRows = new ArrayList<>(optionalPatterns);
         String description = payload.getJsonObject(Utils.PULL_REQUEST).getString(Utils.BODY);
 
         try (Scanner scanner = new Scanner(description)) {
-            while (scanner.hasNextLine() && !requiredRows.isEmpty()) {
+            while (scanner.hasNextLine() || (!requiredRows.isEmpty() && !optionalRows.isEmpty())) {
```
xstefank, line 46: This doesn't seem correct. You want to iterate until there is still check in *either* requiredRows OR optionalRows.

tyr-runner/src/test/java/org/jboss/tyr/check/DescriptionCheckTest.java
```diff
@@ -0,0 +1,117 @@
+/*
+ * Copyright 2019 Red Hat, Inc, and individual contributors.
```
xstefank, line 2: 2021?

tyr-runner/src/test/java/org/jboss/tyr/check/DescriptionCheckTest.java
```diff
@@ -0,0 +1,117 @@
...
+
+    @Test
+    public void checkSimpleRegexMatch() {
+        Assertions.assertNull(descriptionCheck.check(TestUtils.TEST_PAYLOAD), "Cannot match valid description");
+    }
+
+    @Test
+    public void checkSimpleRegexNonMatch() {
+        row.setPattern(Pattern.compile("can't.*match.*this"));
+       Assertions.assertNotNull(descriptionCheck.check(TestUtils.TEST_PAYLOAD), "Matched invalid description");
```
xstefank, line 58: also test the message is correct.

xstefank (changes requested):

tyr-runner/src/test/java/org/jboss/tyr/check/DescriptionCheckTest.java
petrberan: This class was created back in 2019, originally it was requiredRowsCheck, I added 2 test and rename it to better reflect it's content. Should I change the year anyway?

xstefank: yes, the copyright then should be 2019-2021.

tyr-runner/src/main/java/org/jboss/tyr/check/DescriptionCheck.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Red Hat, Inc, and individual contributors.
+ * Copyright 2021 Red Hat, Inc, and individual contributors.
```
xstefank, line 2:
```suggestion
 * Copyright 2019-2021 Red Hat, Inc, and individual contributors.
```

tyr-runner/src/main/java/org/jboss/tyr/check/DescriptionCheck.java
```diff
@@ -48,6 +52,19 @@ public String check(JsonObject payload) {
...
+                for (OptionalPatterns optionalRow : optionalRows){
+                    Matcher preconditionMatcher = optionalRow.getPrecondition().matcher(line);
+                    if (preconditionMatcher.matches()){
+                        Matcher patternMatcher = optionalRow.getPattern().matcher(line);
+                        if (!patternMatcher.matches()){
+                            return optionalRow.getMessage();
+                        }
+                        else {
+                            optionalRows.remove(optionalRow);
+                            break;
```
xstefank, line 64: this break is redundant as it's the last statement in the loop.

tyr-runner/src/main/java/org/jboss/tyr/model/yaml/Format.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Red Hat, Inc, and individual contributors.
+ * Copyright 2019-2021 Red Hat, Inc, and individual contributors.
```
xstefank, line 2: this is only applied when you actually change the file.

tyr-runner/src/main/java/org/jboss/tyr/model/yaml/OptionalPatterns.java
```diff
@@ -0,0 +1,49 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.tyr.model.yaml;
+
+import java.util.regex.Pattern;
+
+public class OptionalPatterns {
```
xstefank, line 20:
```suggestion
public class OptionalPattern {
```

tyr-runner/src/main/resources/format-example.yaml
```diff
@@ -10,17 +10,18 @@ format:
...
+        pattern: "\\[WFLY-\\d+\\]"
+        message: "JIRA issues can contain only numbers"
 
   commands:
     AddUserCommand: "ok\\s+to\\s+test$"
     RetestCommand: "retest\\s+this\\s+please$"
     RetestFailedCommand: "retest\\s+failed\\s+please$"
-
-  CI:
-    - TeamCity
```
xstefank, line 26: why is this removed?

xstefank (changes requested):

tyr-runner/src/main/java/org/jboss/tyr/check/DescriptionCheck.java
petrberan: Isn't the same case on line 52 too?

xstefank: my bad, the intent is to finish when you find a matching row. Resolved.

tyr-runner/src/main/java/org/jboss/tyr/check/DescriptionCheck.java
```diff
@@ -48,6 +52,19 @@ public String check(JsonObject payload) {
...
                 }
+                for (OptionalPatterns optionalRow : optionalRows){
+                    Matcher preconditionMatcher = optionalRow.getPrecondition().matcher(line);
+                    if (preconditionMatcher.matches()){
+                        Matcher patternMatcher = optionalRow.getPattern().matcher(line);
+                        if (!patternMatcher.matches()){
+                            return optionalRow.getMessage();
+                        }
+                        else {
+                            optionalRows.remove(optionalRow);
```
petrberan, line 63: Does it make sense to remove optional checks? Or is use case for using one optional check multiple times valid? Technically, same situation happens in case of required checks,

xstefank: yes, this is the concept. Only one occurrence is required for the required row. For optional it doesn't matter. See https://github.com/jboss/tyr/pull/139/files/07db5f9a2d70f81956d81b65fbebdac61cfd3fe6#diff-af594769d38bf7c5dd4aecf304b82eabed030e9fdaeb385b7df6d64f809b4af4R71

xstefank: actually removal of optional check is wrong because we should check all lines that pass precondition. But I won't block the PR on this.

tyr-runner/src/main/resources/format-example.yaml
```diff
@@ -10,17 +10,21 @@ format:
...
 
   commands:
     AddUserCommand: "ok\\s+to\\s+test$"
     RetestCommand: "retest\\s+this\\s+please$"
     RetestFailedCommand: "retest\\s+failed\\s+please$"
-
+    
   CI:
-    - TeamCity
+    - TeamCity
```
xstefank, line 30: also won't block on this but beware of this kind of formatting changes in the future. They are often regarded as just unnecessary noise.

xstefank (approved):
We will follow up with a small change in the removal of optional checks.

PR #140, opened 2021-01-15
Refactor DesriptionCheck to use iterators instead of breaks and to no…

xstefank:
…t remove optional checks


xstefank:
@petrberan review please.

petrberan (approved):

PR #149, opened 2021-02-14
[#144] Extract tyr-core from the tyr-runner so it can be reused

### LINKED ISSUE #144: Extract tyr-core from the tyr-runner so it can be reused

xstefank:
Resolves #144 

Some of the dependencies can probably be omitted but we can do that in a follow-up PR.


xstefank:
@petrberan, @xvanick1 please review.

petrberan (approved):

xvanick1 (approved):

PR #193, opened 2021-07-22
Fixed close client warning.

xjusko:
Fixes https://github.com/jboss/tyr/issues/192


tyr-core/src/main/java/org/jboss/tyr/ci/TeamCity.java
```diff
@@ -81,6 +82,13 @@ public void init() {
         this.jsonb = JsonbBuilder.create();
     }
 
+    @PreDestroy
```
xstefank, line 85: I just wonder if this shouldn't rather be in the explicit close method since the client is created in explicit init method but we can change that later if needed.

xstefank (approved):

Issue #186, opened 2021-04-19
Add check for the required number of commits in the PR


xstefank:
Many projects require only one commit per PR. We should add a check that verifies user-defined number commit s per PR. This should also support a range regex if needed. So the config could look like this:

```
numberOfCommits: 1
```
or 
```
numberOfCommits: 1..5
```

xvanick1:
> Many projects require only one commit per PR. We should add a check that verifies user-defined number commit s per PR. This should also support a range regex if needed. So the config could look like this:
> 
> ```
> numberOfCommits: 1
> ```
> 
> or
> 
> ```
> numberOfCommits: 1..5
> ```

Do you have any bounds for the maximum of commits?

xstefank:
@xvanick1 not really, I think possitive `short` range should be enough. If not, we will change it later.

PR #196, opened 2021-07-27
Check allowed quantity of commits in PR

xvanick1:
New feature allowing users to define how many commits are allowed in PRs.

Closes issue #186


tyr-core/src/test/java/org/jboss/tyr/check/CommitsQuantityCheckTest.java
```diff
@@ -0,0 +1,110 @@
...
+    public void testNullCommitsMessageParameter() throws InvalidPayloadException {
+        commitsQuantity.setMessage(null);
+        commitsQuantity.setQuantity("5");
+        commitsQuantityCheck.quantityCheck(commitsQuantity);
+
+        Assertions.assertEquals(CommitsQuantityCheck.DEFAULT_MESSAGE, commitsQuantityCheck.check(TestUtils.TEST_PAYLOAD));
+    }
+
+    @Test
+    public void testCheckValueRegexMatchPayloadMatch() throws InvalidPayloadException {
```
xstefank, line 62: you shouldn't mentioned tested internals in the test names, what if we change valueRegex field in future?

tyr-core/src/test/java/org/jboss/tyr/check/CommitsQuantityCheckTest.java
```diff
@@ -0,0 +1,110 @@
...
+        Assertions.assertNull(commitsQuantityCheck.check(TestUtils.TEST_PAYLOAD));
+    }
+
+    @Test
+    public void testCheckValueRegexMatchPayloadMismatch() throws InvalidPayloadException {
+        commitsQuantity.setQuantity("2");
+        commitsQuantity.setMessage("Test: This quantity of commits is not allowed!");
+        commitsQuantityCheck.quantityCheck(commitsQuantity);
+
+        Assertions.assertEquals(commitsQuantity.getMessage(), commitsQuantityCheck.check(TestUtils.TEST_PAYLOAD));
```
xstefank, line 75: maybe test the string matches here directly because if the commitsQuantity would change the message internally your test wouldn't catch it.

xstefank (changes requested):

tyr-core/src/main/java/org/jboss/tyr/check/CommitsQuantityCheck.java
```diff
@@ -0,0 +1,78 @@
...
+    @Inject
+    GitHubService gitHubService;
+
+    private String quantity;
+    private String message;
+
+    private Matcher valueMatcher;
+    private Matcher rangeMatcher;
+
+    private int g1;
```
xstefank, line 47: nope, we don't accept names for variables like this.

tyr-core/src/main/java/org/jboss/tyr/check/CommitsQuantityCheck.java
```diff
@@ -0,0 +1,78 @@
...
+        quantity = commitsQuantity.getQuantity();
+        message = (commitsQuantity.getMessage() != null) ? commitsQuantity.getMessage() : DEFAULT_MESSAGE;
+
+        valueMatcher = valueRegex.matcher(quantity);
+        rangeMatcher = rangeRegex.matcher(quantity);
+
+        if (!(valueMatcher.matches() || rangeMatcher.matches()))
+            throw new IllegalArgumentException("Unrecognized commit range syntax. Use specific number, or range in format 1-20. Allowed upper boundary is 100.");
+
+        if (rangeMatcher.matches()){
```
xstefank, line 63: you are duplicating information in matchers and whatever g1, g2 stands for. matchers don't need to be in fields, you know which case (single number or range) you have depending on whether you set both bounds or only one.

tyr-core/src/main/java/org/jboss/tyr/check/CommitsQuantityCheck.java
```diff
@@ -0,0 +1,78 @@
...
+        if (commitsQuantity == null || commitsQuantity.getQuantity() == null) {
+            throw new IllegalArgumentException("Quantity was not set");
+        }
+        quantity = commitsQuantity.getQuantity();
+        message = (commitsQuantity.getMessage() != null) ? commitsQuantity.getMessage() : DEFAULT_MESSAGE;
+
+        valueMatcher = valueRegex.matcher(quantity);
+        rangeMatcher = rangeRegex.matcher(quantity);
+
+        if (!(valueMatcher.matches() || rangeMatcher.matches()))
```
xstefank, line 60: maybe better to rewrite it as if (valueMatcher) {...} else if (rangeMatcher) {...} else {throw not matched} so you don't need to parse regexes and potentially stop faster.

xstefank (changes requested):

tyr-core/src/main/java/org/jboss/tyr/check/CommitsQuantityCheck.java
```diff
@@ -0,0 +1,85 @@
...
+    private Integer parsedRangeBottomBoundary;
+    private Integer parsedRangeUpperBoundary;
+
+    public void initCheck(CommitsQuantity commitsQuantity) {
+        if (commitsQuantity == null || commitsQuantity.getQuantity() == null) {
+            throw new IllegalArgumentException("Quantity was not set");
+        }
+
+        String quantity = commitsQuantity.getQuantity();
+        message = (commitsQuantity.getMessage() != null) ? commitsQuantity.getMessage() : DEFAULT_MESSAGE;
```
xstefank, line 52: parentheses should not be needed

tyr-core/src/main/java/org/jboss/tyr/check/CommitsQuantityCheck.java
```diff
@@ -0,0 +1,85 @@
...
+
+        Matcher valueMatcher = valueRegex.matcher(quantity);
+        Matcher rangeMatcher = rangeRegex.matcher(quantity);
```
xstefank, line 55: it would be better to inline these directly to the if conditions below so we might avoid creating rangeMatcher object.

tyr-core/src/test/java/org/jboss/tyr/check/CommitsQuantityCheckTest.java
```diff
@@ -0,0 +1,108 @@
...
+        Assertions.assertNull(commitsQuantityCheck.check(TestUtils.TEST_PAYLOAD));
+    }
+
+    @Test
+    public void testPayloadMismatch() throws InvalidPayloadException {
+        commitsQuantity.setQuantity("2");
+        commitsQuantity.setMessage("Test: This quantity of commits is not allowed!");
+        commitsQuantityCheck.initCheck(commitsQuantity);
+
+        Assertions.assertEquals("Test: This quantity of commits is not allowed!", commitsQuantityCheck.check(TestUtils.TEST_PAYLOAD));
```
xstefank, line 75: better to take repeated strings in a separate variable

xstefank (approved):
Minor comments, merging. Thanks!

PR #204, opened 2021-08-19
[Issue #194]Added case-insensitive modifier to tyr-format.yaml regex.

### LINKED ISSUE #194: tyr-format.yaml should work with capitalized words
In https://github.com/jboss/tyr/blob/master/.github/tyr-format.yaml#L9 we list everything in lowercase but GH works with uppercase too.

xjusko:
fixes #194


xstefank (approved):
I was a little skeptical about fiXes or fixeS, but apparently it works - https://github.com/xstefank/test-repo/pull/75

